PR #153, opened 2023-08-22
Use less inflammatory language for `readLine()` protection

nahsra:


core-codemods/src/main/resources/io/codemodder/codemods/LimitReadlineCodemod/description.md
```diff
@@ -1,4 +1,4 @@
-This change hardens all [`BufferedReader#readLine()`](https://docs.oracle.com/javase/8/docs/api/java/io/BufferedReader.html#readLine--) operations against attack.
+This change hardens all [`BufferedReader#readLine()`](https://docs.oracle.com/javase/8/docs/api/java/io/BufferedReader.html#readLine--) operations against catastrophic conditions.
```
gilday, line 1: I like "memory exhaustion", because it's more direct. The next sentence goes on to explain how the memory gets exhausted.

core-codemods/src/main/resources/io/codemodder/codemods/LimitReadlineCodemod/description.md
gilday, line 1: > A stream influenced by an attacker could keep providing bytes until the JVM runs out of memory.

I think we can introduce the idea of this problem escalating to an attack in certain situations by using some transition word to indicate that memory exhaustion leads to attacks and by elaborating on "stream influence by an attack" for the less app sec literate developers.

> Furthermore, a stream of data provided by an untrusted source could lead to a denial of service attack, as attackers can provide an infinite stream of bytes until the process runs out of memory.

gilday (approved):

PR #154, opened 2023-08-24
feature/dostrow replacegenericsinvar codemod

danielostrow:
This codemod is for replacing generics in var creation.

### DIFF
--- a/core-codemods/src/main/java/io/codemodder/codemods/ReplaceGenericsInVarCodemod.java
+++ b/core-codemods/src/main/java/io/codemodder/codemods/ReplaceGenericsInVarCodemod.java
@@ -1,0 +1,4 @@
+package io.codemodder.codemods;
+
+public class ReplaceGenericsInVarCodemod {
+}
--- a/core-codemods/src/main/java/io/codemodder/codemods/ReplaceGenericsInVarCodemod.java
+++ b/core-codemods/src/main/java/io/codemodder/codemods/ReplaceGenericsInVarCodemod.java
@@ -1,4 +1,63 @@
 package io.codemodder.codemods;
 
-public class ReplaceGenericsInVarCodemod {
+import com.github.javaparser.ast.CompilationUnit;
+import com.github.javaparser.ast.body.VariableDeclarator;
+import com.github.javaparser.ast.expr.Expression;
+import com.github.javaparser.ast.expr.StringLiteralExpr;
+import com.github.javaparser.ast.expr.VariableDeclarationExpr;
+import io.codemodder.*;
+import io.codemodder.providers.sarif.semgrep.SemgrepScan;
+import javax.inject.Inject;
+
+@Codemod(
+    id = "pixee:java/replace-generics-in-var",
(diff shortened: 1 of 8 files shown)


core-codemods/src/main/java/io/codemodder/codemods/ReplaceGenericsInVarCodemod.java
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  @Override
+  public boolean onResultFound(
+      CodemodInvocationContext context,
+      CompilationUnit cu,
+      VariableDeclarationExpr variableDeclarationExpr,
+      com.contrastsecurity.sarif.Result result) {
+
+    for (VariableDeclarator variableDeclarator : variableDeclarationExpr.getVariables()) {
```
nahsra, line 31: I'm not sure this is right, but it doesn't _feel_ right. You can define multiple vars in a single declaration statement:

```
List<String> a = new ArrayList<String>(), b = new ArrayList<>();
```

It seems like the Semgrep is probably identifying one, but we're acting on all of them. It may be correct, because the left hand of the expression may make it redundant on all the right hand initializers, but I'm too tired to think through that right now. Anyhow, it definitely warrants a comment to discuss this.

core-codemods/src/main/java/io/codemodder/codemods/ReplaceGenericsInVarCodemod.java
```diff
@@ -0,0 +1,63 @@
...
+
+                // Remove wrapped quotes
+                String formatted =
+                    cleanedInitializer
+                        .toString()
+                        .substring(1, cleanedInitializer.toString().length() - 1);
+
+                variableDeclarator.setInitializer(formatted);
+              });
+    }
```
nahsra, line 59: I don't like the way this code is written because the place where we `return true` actually has no idea if any changes were made. It's tempting to put a lot of code into those `ifPresent()` and `ifEmpty()` chains, but it's tough get any state to escape from them.

nahsra (changes requested):

core-codemods/src/main/resources/io/codemodder/codemods/replace-generics-in-var.yaml
```diff
@@ -0,0 +1,18 @@
+rules:
+  - id: replace-generics-in-var
+    patterns:
+      - pattern-either:
```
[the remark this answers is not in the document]
danielostrow: I'll make the changes.

Having `$X` in type covers cases when someone declares `Map` or `List`. The initializer will be a different value for the type and I needed a separate variable `$X` to read it. There was no organization in variable choices, I just chose different variables. I will correct that.

I will limit the rule to 2-length generics, however I do not see the negatives in including more than 2 besides a goofy looking rule.

nahsra:
I'm closing this PR. Feel free to re-open when you're closer to a mergable state. Until then, it clutters about my view of the work that is "close to the main branch".

PR #156, opened 2023-08-29
Merged LLM test code to create shared test infrastructure

nahsra:
This resulted in better tests also for all the existing deterministic codemods, as there were good common sense checks only appearing in the LLM test code.


framework/codemodder-testutils/src/main/java/io/codemodder/testutils/CodemodTestMixin.java
```diff
@@ -117,6 +156,11 @@ private void verifyCodemod(
     List<CodeTFChangesetEntry> changeset2 = result2.getChangeset();
     assertThat(changeset2.size(), is(0));
     String transformedAgainJavaCode = Files.readString(pathToJavaFile);
+    String expectedCode = Files.readString(after);
     assertThat(transformedAgainJavaCode, equalTo(expectedCode));
```
vsomayaji, line 160: This bit will break if the LLM made a change that wasn't an exact match (i.e. `pathToJavaFile` didn't match `after` before the codemod was re-run, so it won't match it now).

framework/codemodder-testutils/src/main/java/io/codemodder/testutils/CodemodTestMixin.java
```diff
@@ -37,22 +47,39 @@ default void it_verifies_codemod(@TempDir final Path tmpDir) throws IOException
                 })
             .toList();
 
-    Path testDir = Path.of("src/test/resources/" + testResourceDir);
-    verifyCodemod(codemod, tmpDir, testDir, dependencies);
+    for (Path path : allBeforeFiles) {
+      verifyCodemod(codemod, tmpDir, path.getParent(), path, dependencies);
+    }
```
vsomayaji, line 52: One thing I miss about the old way: all of the examples are now grouped under the `it_verifies_codemod` test, and one failed example will keep the remaining examples from being run.

Screenshot 2023-08-30 at 8 09 48 AM

The old way:

Screenshot 2023-08-30 at 8 19 02 AM

nahsra: That's true. I can possibly get that back.

nahsra: Re-added!

framework/codemodder-testutils/src/main/java/io/codemodder/testutils/CodemodTestMixin.java
```diff
@@ -116,7 +149,39 @@ private void verifyCodemod(
     CodeTFResult result2 = executor2.execute(List.of(pathToJavaFile));
     List<CodeTFChangesetEntry> changeset2 = result2.getChangeset();
     assertThat(changeset2.size(), is(0));
-    String transformedAgainJavaCode = Files.readString(pathToJavaFile);
+
+    verifyRetransformedCode(before, after, pathToJavaFile);
```
vsomayaji, line 153: This'll work, but I think we'll make an unnecessary LLM call. Let's say the first codemod run changes `pathToJavaFile` to `after*` and the LLM says it's acceptable. After the second codemod run, if nothing has changed, `pathToJavaFile` will still be `after*` and we'll call the LLM again.

Instead, if we save the contents of `pathToJavaFile` to `transformedCode` before the second codemod run and then `assertThat(transformedAgainCode, is(transformedCode))`, that should work for both types of codemods.

framework/codemodder-testutils/src/main/java/io/codemodder/testutils/CodemodTestMixin.java
```diff
@@ -1,32 +1,50 @@
...
+import org.junit.jupiter.api.function.ThrowingConsumer;
 import org.junit.jupiter.api.io.TempDir;
 
 /** The basic tests for codemods. */
 public interface CodemodTestMixin {
 
-  @Test
-  default void it_verifies_codemod(@TempDir final Path tmpDir) throws IOException {
+  @TestFactory
+  @EnabledIfEnvironmentVariable(named = "CODEMODDER_OPENAI_API_KEY", matches = ".+")
```
vsomayaji, line 32: Don't we only want this annotation on the LLM-verifying tests?

nahsra: True, fixed, derp

PR #159, opened 2023-08-29
Add dependency change messages

nahsra:
This change will add facts to the CodeTF change description that allow more advanced storytelling on dependency adds.


aldrinleal (approved):
Lgtm

PR #160, opened 2023-08-30
Work on files outside of traditional source directories

nahsra:
This change allows us to transform files that are not inside traditional source directories. There appears to be no downside tradeoffs, except that they will have weaker symbol resolution capabilities.

### DIFF
--- a/framework/codemodder-base/src/main/java/io/codemodder/CLI.java
+++ b/framework/codemodder-base/src/main/java/io/codemodder/CLI.java
@@ -390,9 +390,7 @@ final class CLI implements Callable<Integer> {
     if (parameterStrings == null || parameterStrings.isEmpty()) {
       return List.of();
     }
-    return parameterStrings.stream()
-        .map(ParameterArgument::fromNameValuePairs)
-        .collect(Collectors.toUnmodifiableList());
+    return parameterStrings.stream().map(ParameterArgument::fromNameValuePairs).toList();
   }
 
   private static final int SUCCESS = 0;
@@ -412,7 +410,7 @@ final class CLI implements Callable<Integer> {
           "**/web.xml");
 
   private static final List<String> defaultPathExcludes =
-      List.of("**/test/**", "**/target/**", "**/build/**");
+      List.of("**/test/**", "**/target/**", "**/build/**", "**/.mvn/**");
 
   private static final Logger log = LoggerFactory.getLogger(CLI.class);
 }
--- a/framework/codemodder-base/src/main/java/io/codemodder/CLI.java
+++ b/framework/codemodder-base/src/main/java/io/codemodder/CLI.java
@@ -312,7 +312,6 @@ final class CLI implements Callable<Integer> {
       JavaParser javaParser = javaParserFactory.create(sourceDirectories);
       CachingJavaParser cachingJavaParser = CachingJavaParser.from(javaParser);
       for (CodemodIdPair codemod : codemods) {
-        System.out.println("Running codemod: " + codemod.getId());
         CodemodExecutor codemodExecutor =
             new DefaultCodemodExecutor(
                 projectPath,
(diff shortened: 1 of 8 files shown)


framework/codemodder-base/src/main/java/io/codemodder/CLI.java
```diff
@@ -166,19 +168,17 @@ enum OutputFormat {
...
-                    f -> !Files.isSymbolicLink(f)) // could cause infinite loop if we follow links
-                .sorted()
-                .toList());
+      try (Stream<Path> paths = Files.walk(projectDir)) {
+        paths
+            .filter(Files::isRegularFile)
+            .filter(p -> !Files.isSymbolicLink(p)) // could cause infinite loop if we follow links
+            .filter(p -> includesExcludes.shouldInspect(p.toFile()))
+            .sorted()
+            .forEach(allFiles::add);
```
ryandens, line 179:
```suggestion
      final List<Path> allFiles;
      try (Stream<Path> paths = Files.walk(projectDir)) {
        allFiles = paths
            .filter(Files::isRegularFile)
            .filter(p -> !Files.isSymbolicLink(p)) // could cause infinite loop if we follow links
            .filter(p -> includesExcludes.shouldInspect(p.toFile()))
            .sorted()
            .toList();
```

Lets use `Stream.toList()` inside the try-with-resources rather than creating a list and adding to it inside a `forEach`

ryandens: chatgpt prompts to get this, because I looked at it and thought it was a bit odd:

> how to use Files.walk java api properly  

outputted something very similar to what you have here

> How can I see Files.walk and the Stream Collect API

Gave what i suggested

ryandens (changes requested):
Overall LGTM but the integration test is failing so not sure what changes are needed to get it working properly

framework/codemodder-base/src/test/java/io/codemodder/CLITest.java
```diff
@@ -46,11 +46,12 @@ void setup(final @TempDir Path tmpDir) throws IOException {
         fooJavaFile,
         "import com.acme.util.Bar; class Foo {private var bar = new Bar();}".getBytes());
     Files.write(barJavaFile, "public class Bar {}".getBytes());
+
+    /*
+     * Only add module2 to the discovered source directories. This will help prove that the module2 files can still be seen and changed, even if we couldn't locate it as a "source directory".
```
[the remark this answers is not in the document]
nahsra: Yes, derp

nahsra:
> Overall LGTM but the integration test is failing so not sure what changes are needed to get it working properly

Amazing, this produced 2 unexpected changes, both positive. First, it allowed us to modify the root `pom.xml` and inject the dependencies we need all the way up, which is what was happening locally before a few weeks ago when we were facing the last-second `pom-operator` issues. 

Second, it found new files to change in the `.mvn` directory, which I guess is some dynamically generated stuff, maybe for `mvnw`? I added a change to make sure that's excluded by default.

image

nahsra:
> > Overall LGTM but the integration test is failing so not sure what changes are needed to get it working properly
> 
> Amazing, this produced 2 unexpected changes, both positive. First, it allowed us to modify the root `pom.xml` and inject the dependencies we need all the way up, which is what was happening locally before a few weeks ago when we were facing the last-second `pom-operator` issues.
> 
> Second, it found new files to change in the `.mvn` directory, which I guess is some dynamically generated stuff, maybe for `mvnw`? I added a change to make sure that's excluded by default. I'm adding tests for this.
> 
> image

Too bad: ignoring the `.mvn` directory caused us to not alter the root `pom.xml` file, so looks that will continue to be imperfect. But, good validation that we really will scan Java files anywhere now.

nahsra:
Ready for re-review

ryandens (approved):
:shipit: Nice!

PR #162, opened 2023-09-01
Add contextual flag to dependency change descriptions

nahsra:
This change adds a marker to CodeTF to show that the changes for dependencies are contextual, and higher value add than others may be.

### DIFF
--- a/plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/MavenProvider.java
+++ b/plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/MavenProvider.java
@@ -21,15 +21,7 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Objects;
-import java.util.Optional;
-import java.util.Set;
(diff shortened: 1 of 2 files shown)


plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/MavenProvider.java
```diff
@@ -282,8 +274,17 @@ private CodeTFChangesetEntry getChanges(
     String relativePomPath = projectDir.relativize(pomDocumentPath).toString();
 
     String description = dependencyDescriptor.create(newDependency);
+    final Map<String, String> properties;
+    if (description != null && !description.isBlank()) {
+      /*
+       * Tell downstream consumers that this is a change based on surrounding context.
+       */
+      properties = Map.of("contextual", "true");
```
bdoyal, line 282: `MavenProvider` oriented codemods are the ones that list the facts for their dependency changes, correct? Does that mean secure random will have these facts as well?  We will have to sort out how to mix the two in cases where we need to in the enrichment service in cases where we want to show facts and enrich the description.  In the near term it will help or have those without facts (like secure random?) provide empty descriptions.

nahsra: `Random` -> `SecureRandom` changes will not have these facts, because it doesn't add a dependency. The `MavenProvider` has this info hardcoded for now, and if downstream consumers want to correlate/replace this, fine with me.

bdoyal (approved):

PR #163, opened 2023-09-04
Refactored `LLMAssistedCodemod` down

nahsra:
The `LLMAssistedCodemod` was a god-class, and this is the first step towards trying to support many different types of LLM-assisted codemod architectures:

* the name was too broad, since many codemods of various shapes will be LLM-assisted
* it had a lot of inner classes will be valuable to other types, so they were extracted
* `CodemodInvocationContextFile` renamed to a package-protected `FileDescription` type

### DIFF
--- a/framework/codemodder-base/src/main/java/io/codemodder/CodemodResources.java
+++ b/framework/codemodder-base/src/main/java/io/codemodder/CodemodResources.java
@@ -1,0 +1,43 @@
+package io.codemodder;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.UncheckedIOException;
+import java.nio.charset.StandardCharsets;
+import java.util.MissingResourceException;
+
+/** A utility class for accessing a codemod's resources in it's "default location" the classpath. */
+public final class CodemodResources {
+
+  private CodemodResources() {}
+
+  /**
+   * Returns a class resource as a {@code String}.
+   *
+   * <p>The absolute name of the class resource is of the following form:
+   *
+   * <blockquote>
+   *
+   * {@code /modifiedPackageName/className/relativeName}
+   *
+   * </blockquote>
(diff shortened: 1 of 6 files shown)


framework/codemodder-base/src/main/java/io/codemodder/CodemodResources.java
```diff
@@ -0,0 +1,43 @@
...
+   * <blockquote>
+   *
+   * {@code /modifiedPackageName/className/relativeName}
+   *
+   * </blockquote>
+   *
+   * Where the {@code modifiedPackageName} is the package name of this object with {@code '/'}
+   * substituted for {@code '.'}.
+   *
+   * @param relativeName The relative name of the resource.
```
vsomayaji, line 28: Maybe: add `@param type`?

vsomayaji (approved):
Much cleaner! 👍🏾 

A few comments, but nothing major.

framework/codemodder-base/src/main/java/io/codemodder/CodemodResources.java
```diff
@@ -0,0 +1,43 @@
+package io.codemodder;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.UncheckedIOException;
+import java.nio.charset.StandardCharsets;
+import java.util.MissingResourceException;
+
+/** A utility class for accessing a codemod's resources in it's "default location" the classpath. */
```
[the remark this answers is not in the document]
nahsra: Ill fix this in the next PR

framework/codemodder-base/src/main/java/io/codemodder/CodemodResources.java
nahsra: Same

PR #173, opened 2023-09-14
🔥 stop publishing container image of codemodder-java as downstream repos now consume it as a Maven artifact

ryandens:


nahsra (approved):
lgtm

PR #174, opened 2023-09-14
:bug: fix issue where github-actions[bot] can no longer push to main

ryandens:


version.txt
```diff
@@ -1 +1 @@
-0.67.2-SNAPSHOT
```
nahsra, line 1: I added this to match the current next version: https://search.maven.org/search?q=codemodder

nahsra (approved):

version.txt
nahsra: Maybe this wouldn't have mattered but I didn't want a confusing wrong version sitting in there

ryandens: i think it would have just gotten overwritten on the next release, but agreed it was good to update.

PR #180, opened 2023-09-22
Update `module-info.java` when needed

nahsra:
This PR adds support for updating `module-info.java` when it's detected. This solution isn't complete in offering JPMS support. There are other ways of configuring `module-info.java`, like using [moditect](https://github.com/moditect/moditect). However, this solves for a real user issue. 

If dependencies want to take advantage of being injected, they must offer a "module name" (which is usually just the core package that library offers.)


nahsra:
Holding this PR in draft until `java-security-toolkit` is modularized and we can test it.

PR #183, opened 2023-09-26
Move from `String`-based charset lookup to `StandardCharsets`

nahsra:
This codemod is straightforward but it's notable in that it attempts to "clean up" the exception handling of the method, which may need changes after the code change. 

Compare the `Test.java.before/after` to see the types of code changes this now supports. Happily, the type resolution provided by JavaParser means this will generally be correct a lot. I think this is safe enough to move to `DefaultCodemods` -- meaning it will be right 95% of the time, and when it's not right, it's an easy fix -- but happy to hear your opinion.

### DIFF
--- a/core-codemods/src/test/resources/switch-to-standard-charsets/Test.java.after
+++ b/core-codemods/src/test/resources/switch-to-standard-charsets/Test.java.after
@@ -1,0 +1,47 @@
+package com.acme.testcode;
+
+import java.io.UnsupportedEncodingException;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+
+final class Test {
+
+   void foo(String str) throws UnsupportedEncodingException {
+       str.getBytes(StandardCharsets.UTF_8);
+       str.getBytes(StandardCharsets.UTF_16);
+       str.getBytes(StandardCharsets.UTF_16LE);
+       str.getBytes(StandardCharsets.UTF_16BE);
+       str.getBytes(StandardCharsets.US_ASCII);
+       str.getBytes(StandardCharsets.ISO_8859_1);
+       str.getBytes("unknown charset");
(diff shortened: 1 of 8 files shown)


core-codemods/src/test/resources/switch-to-standard-charsets/Test.java.after
```diff
@@ -0,0 +1,56 @@
...
+final class Test {
+
+   void foo(String str) throws UnsupportedEncodingException {
+       str.getBytes(StandardCharsets.UTF_8);
+       str.getBytes(StandardCharsets.UTF_16);
+       str.getBytes(StandardCharsets.UTF_16LE);
+       str.getBytes(StandardCharsets.UTF_16BE);
+       str.getBytes(StandardCharsets.US_ASCII);
+       str.getBytes(StandardCharsets.ISO_8859_1);
+       str.getBytes("unknown charset");
```
nahsra, line 16: Because this call is still here after the code change (because we can't safely translate this to `StandardCharsets` constant), we can't remove the `UnsupportedEncodingException` from the method signature.

core-codemods/src/test/resources/switch-to-standard-charsets/Test.java.after
```diff
@@ -0,0 +1,56 @@
...
+       Object o = new Object();
+       o.getBytes("UTF-8"); // this is not a string, so shouldn't act on it
+   }
+
+   void bar(String s) {
+     Charset c = StandardCharsets.UTF_8;
+     byte[] b = s.getBytes(StandardCharsets.UTF_16LE);
+     Charset.forName("unknown");
+   }
+
```
nahsra, line 28: In `removeTryBlock`, we show the design decision to leave the existing code in a `{ }` block. We do this because we don't want to change the scope of any of the existing variables inside the scope, which would require a more complicated analysis to measure correctness for.

PR #196, opened 2023-10-13
Make sure scanners don't scan files that aren't inspected

nahsra:
When invoking Semgrep and PMD, scanners were scanning _all files_, even if codemods weren't being asked to scan _all_ files. This change gives all the providers all the "include/exclude" configuration so they can parameterize their scans accordingly.

For some context, this PR increases the speedup of some large projects from 10 minutes to 1 minute, when scanning one file.

### DIFF
--- a/framework/codemodder-base/src/main/java/io/codemodder/CodemodLoader.java
+++ b/framework/codemodder-base/src/main/java/io/codemodder/CodemodLoader.java
@@ -20,43 +20,12 @@ public final class CodemodLoader {
   private final List<CodemodIdPair> codemods;
 
   public CodemodLoader(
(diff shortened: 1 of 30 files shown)


framework/codemodder-base/src/main/java/io/codemodder/CodemodLoader.java
```diff
@@ -19,44 +19,13 @@ public final class CodemodLoader {
 
   private final List<CodemodIdPair> codemods;
```
nahsra, line 21: Removed all these confusing intermediate constructors as it made working through them tough.

dhafley (approved):

PR #197, opened 2023-10-13
Integration test framework - initial approach

arturoespinoza15:
Summary
We want to create a test framework for codemods, we want to verify that the behavior of a test application don't change after code transfrmation

What changed:
- Dockerfile to create codemodder base image was added with the required dependencies to run the tool
- Dockerfile to create the different test project images was added
- Dockerfile used to create a container to run the test application was added
- `CodemodIntegrationTestMixin` utility class created to manage the containers and tests creation
- `TestApplicationRequestUtil` utility class created to perform request to the tests endpoints
- Integration tests added for `verbose-request-mapping` and `move-switch-default-last` codemods


core-codemods/src/test/java/io/codemodder/codemods/integration/util/CodemodIntegrationTestMixin.java
```diff
@@ -0,0 +1,67 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.Arrays;
+import java.util.stream.Stream;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.DynamicTest;
+import org.junit.jupiter.api.TestFactory;
+import org.testcontainers.containers.GenericContainer;
+
+public class CodemodIntegrationTestMixin {
```
nahsra, line 13: Need JavaDoc for all `public` types and methods. Are we sure this couldn't be an `interface`? If it was, we could have a test the implemented multiple mixins.

core-codemods/src/test/java/io/codemodder/codemods/integration/util/CodemodIntegrationTestMixin.java
```diff
@@ -0,0 +1,67 @@
...
+                        doRequest(transformedContainerEndpointURL, httpVerb);
+
+                    assertThat(originalCodeResponse).isEqualTo(expectedResponse);
+                    assertThat(transformedCodeResponse).isEqualTo(expectedResponse);
+                  });
+            });
+  }
+
+  @AfterAll
+  public static void tearDown() {
```
nahsra, line 63: This doesn't need to be `public` with JUnit 5

core-codemods/src/test/java/io/codemodder/codemods/integration/util/DockerContainerFactory.java
```diff
@@ -0,0 +1,25 @@
+package io.codemodder.codemods.integration.util;
+
+import java.nio.file.Path;
+import org.testcontainers.containers.GenericContainer;
+import org.testcontainers.images.builder.ImageFromDockerfile;
+
+public class DockerContainerFactory {
```
nahsra, line 7: JavaDoc for everything public in this type

core-codemods/src/test/java/io/codemodder/codemods/integration/util/TestApplicationRequestUtil.java
```diff
@@ -0,0 +1,47 @@
+package io.codemodder.codemods.integration.util;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.net.HttpURLConnection;
+import java.net.URL;
+
```
nahsra, line 8: Why don't we call this type `Requests`? The `ThingUtil` pattern has fallen out of favor.

core-codemods/src/test/java/io/codemodder/codemods/integration/util/TestApplicationRequestUtil.java
```diff
@@ -0,0 +1,47 @@
...
+      StringBuffer response = new StringBuffer();
+
+      while ((inputLine = in.readLine()) != null) {
+        response.append(inputLine);
+      }
+      in.close();
+
+      // print result
+      return response.toString();
+    } else {
```
nahsra, line 43: This seems like exceptional behavior, so we should throw an `IOException` or something?

core-codemods/src/test/java/io/codemodder/codemods/integration/util/TestApplicationRequestUtil.java
```diff
@@ -0,0 +1,47 @@
...
+    };
+  }
+
+  private static String sendRequest(final String url, final String httpVerb) throws IOException {
+    URL obj = new URL(url);
+    HttpURLConnection con = (HttpURLConnection) obj.openConnection();
+    con.setRequestMethod(httpVerb);
+    int responseCode = con.getResponseCode();
+    if (responseCode == HttpURLConnection.HTTP_OK) { // success
+      BufferedReader in = new BufferedReader(new InputStreamReader(con.getInputStream()));
```
nahsra, line 32: I think there's some `IOUtils#toString(InputStream)` method we can do to save some some code here.

core-codemods/src/test/java/io/codemodder/codemods/integration/util/TestPropertiesMetadata.java
```diff
@@ -0,0 +1,20 @@
+package io.codemodder.codemods.integration.util;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface TestPropertiesMetadata {
```
nahsra, line 10: This type needs JavaDoc

I think we should consider renaming this to `IntegrationTestMetadata` to be consistent with our normal unit test type.

core-codemods/src/test/resources/test-applications/move-switch-default-last/build.gradle.kts
```diff
@@ -0,0 +1,26 @@
+import org.springframework.boot.gradle.tasks.bundling.BootJar
+
+plugins {
+    java
+    id("org.springframework.boot") version "3.1.4"
+}
+
```
nahsra, line 7: Do we need to version this specifically, or can we get versions from parent or something? Would that cause any issues?

Also, I think we should adjust the `group` here, `ai.pixee.integration`?

core-codemods/src/test/resources/test-applications/move-switch-default-last/gradlew
```diff
@@ -0,0 +1,248 @@
+#!/bin/sh
```
nahsra, line 1: Why do we need our own Gradle here?

nahsra (changes requested):
Looks good! Just a few things. Hoping @ryandens can weigh in on some of the Gradle stuff.

core-codemods/src/test/java/io/codemodder/codemods/integration/util/CodemodIntegrationTestMixin.java
arturoespinoza15: I have spent some time trying figure out  the best way to manage the containers lifecycle in a interface, I am having the following problems:
- I can't instantiate the containers globally since the `codemoId` used in the container creation is available once the annotation parameters are extracted inside the default method.
- Since the @TestFactory annotated method must return a stream of Dynamic test, there is no space to stop the containers just after the tests execution. I have tried to collect the generated tests, execute them and then stop the containers just before return the stream of tests but it fails.
- @Testcontainers and @Container can manage automatically the container lifecycle but I can't declare the containers globally to be annotated and then instantiate them inside the default method in the interface

I will work with the other feedback and I will give it another try later

core-codemods/src/test/java/io/codemodder/codemods/integration/util/TestApplicationRequestUtil.java
arturoespinoza15: renamed

core-codemods/src/test/java/io/codemodder/codemods/integration/util/TestApplicationRequestUtil.java
arturoespinoza15: yes, I am throwing the exception with the status code in the message

core-codemods/src/test/resources/test-applications/move-switch-default-last/gradlew
arturoespinoza15: my bad, we are not using the wrappers to execute comands

core-codemods/src/test/resources/test-applications/move-switch-default-last/build.gradle.kts
arturoespinoza15: yes we can have a common `build.gradle` and all the test projects can take dependencies from there, if we want to enforce a spring boot version in the parent we will have take care of the dependencies used on each project to prevent incompatibility problems

core-codemods/src/test/java/io/codemodder/codemods/integration/util/TestApplicationRequests.java
```diff
@@ -0,0 +1,43 @@
+package io.codemodder.codemods.integration.util;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import org.testcontainers.shaded.org.apache.commons.io.IOUtils;
+
```
nahsra, line 8: If this is meant to be just static methods let's add a no-args private constructor that throws an exception to prevent instantiation

core-codemods/src/test/resources/test-applications/move-switch-default-last/build.gradle.kts
nahsra: Do the versions matter? If we're versioning on the test image (or creating it from scratch every time) then it doesn't really matter.

arturoespinoza15: oh wait, you were referring to the project version(version = "0.0.1-SNAPSHOT") right? I was misunderstanding I though you were talking about the dependencies versions, yes in this case the version property is not required

arturoespinoza15: deleting them

core-codemods/src/test/java/io/codemodder/codemods/integration/util/CodemodIntegrationTestMixin.java
arturoespinoza15: @nahsra I have tried again and now the Mixing class is an interface, I have checked that the containers are properly stopped which was a problem in my first try

nahsra (approved):

PR #199, opened 2023-10-18
Added limits useful for testing

nahsra:
This PR adds the ability to restrict the amount of work done in a codemodder scan. I wish I had these flags useful for testing several times before now, and I think eventually they may be useful for downstream consumers to "limit the work" that a codemodder run can do. 

Hopefully, we'll be following this PR quickly with work to make these parameters obsolete (because removing `CompilationUnit` caching and I/O parallelization will fix everything), but I like having them around in case we ever run into situations where we'd like to performance test locally.

### DIFF
--- a/framework/codemodder-base/src/test/java/io/codemodder/javaparser/DefaultCachingJavaParserTest.java
+++ b/framework/codemodder-base/src/test/java/io/codemodder/javaparser/DefaultCachingJavaParserTest.java
@@ -1,0 +1,48 @@
+package io.codemodder.javaparser;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import com.github.javaparser.ast.CompilationUnit;
+import io.codemodder.SourceDirectory;
+import java.io.IOException;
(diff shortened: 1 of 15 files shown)


framework/codemodder-base/src/test/java/io/codemodder/javaparser/DefaultCachingJavaParserTest.java
```diff
@@ -0,0 +1,48 @@
...
+import com.github.javaparser.ast.CompilationUnit;
+import io.codemodder.SourceDirectory;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
```
nahsra, line 14: We were missing good coverage of the caching JavaParser test. I know this is getting removed soon, but I created this test as part of debugging, so why not keep it in.

PR #206, opened 2023-10-23
Add parallelization

nahsra:
This change introduces parallelizing file analysis within a single codemod's execution. This is done primarily with an eye towards systems with slow file I/O.

The changes really boil down to 3 things: 
* created a per-thread copy of a `JavaParser` because they're not intended to be thread-safe
* made per-file analysis threaded
* synchronized post-file change operations (e.g., changing file updates and updating/invalidating caches)


nahsra:
@drdavella I know you probably don't have too much of an opinion on the Java here. Mainly asking for feedback on the Linux vs. Mac thread count selection guesstimate thing. Open to alternatives.

PR #208, opened 2023-10-24
Meet new threading spec

nahsra:
This PR makes `codemodder-java` meet the [new spec](https://github.com/pixee/codemodder-specs/pull/11/).


drdavella (approved):
I don't have any objections to this but I'm a little surprised to find that you're not benefiting from better performance with more threads. I guess we can do some more investigation later and I'll let you know the results of my own investigation for Python.

PR #210, opened 2023-10-24
Disable caching

nahsra:
This PR introduces all the changes from the long-lived `kill_caching` branch, which itself had the following changes:

* kills the caching of JavaParser models
* refactors PMD/Semgrep to be run before each codemod
* refactors PMD/Semgrep to be run batched to identify which
* allows codemods to be executed with higher priority
* allows a pre-flight check for each codemod to skip execution if unwarranted

Disabling caching alone significantly brought execution time and accuracy down. Therefore, we had to introduce more changes before it could be brought back into `main`.


core-codemods/src/main/java/io/codemodder/codemods/InsecureCookieCodemod.java
```diff
@@ -8,19 +8,17 @@
 import com.github.javaparser.ast.expr.MethodCallExpr;
 import com.github.javaparser.ast.stmt.ExpressionStmt;
 import com.github.javaparser.ast.stmt.Statement;
-import io.codemodder.Codemod;
-import io.codemodder.CodemodInvocationContext;
-import io.codemodder.RegionExtractor;
-import io.codemodder.ReviewGuidance;
-import io.codemodder.RuleSarif;
-import io.codemodder.SarifPluginJavaParserChanger;
+import io.codemodder.*;
```
dhafley, line 11: don't we want to avoid wildcard imports?

core-codemods/src/intTest/java/io/codemodder/integration/WebGoat822Test.java
```diff
@@ -133,7 +133,7 @@ void it_transforms_webgoat_with_codeql() throws Exception {
             .flatMap(Collection::stream)
             .collect(Collectors.toList());
```
nahsra, line 135: There's a filed `JWTVotesEndpoint.java` that has 2 CodeQL findings in it, that we remediate. Because of the new architecture, we're unable to find it because the line numbers changed. This is an expected casualty from the tradeoff we made from caching the JavaParser models. There are solutions to re-capturing this one, but it's not in scope.

core-codemods/src/main/java/io/codemodder/codemods/InsecureCookieCodemod.java
nahsra: It's one of those bikeshed things that we delegate to [spotless](https://github.com/diffplug/spotless) to settle for us

carlosu7 (approved):
I run this change (against trino) and now task ends in 11m 

Same changes are applied

PR #214, opened 2023-10-26
Remove version from Dependency hash code for assertion purposes

carlosu7:
Also I updated assertions library.

### DIFF
--- a/plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/operator/Dependency.java
+++ b/plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/operator/Dependency.java
@@ -70,7 +70,15 @@ public class Dependency {
 
   @Override
   public int hashCode() {
+    return Objects.hash(groupId, artifactId, version, classifier, packaging, scope);
+  }
+
+  public int hashWithoutVersion() {
     return Objects.hash(groupId, artifactId, classifier, packaging, scope);
+  }
+
+  public boolean matchesWithoutConsideringVersion(final Dependency dependency) {
+    return hashWithoutVersion() == dependency.hashWithoutVersion();
   }
 
   public String getGroupId() {
--- a/plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/operator/Dependency.java
+++ b/plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/operator/Dependency.java
@@ -73,10 +73,28 @@ public class Dependency {
     return Objects.hash(groupId, artifactId, version, classifier, packaging, scope);
   }
 
+  /**
+   * Computes the hash code of this object without considering the version.
+   *
+   * This method calculates the hash code based on the values of the group ID, artifact ID,
+   * classifier, packaging, and scope fields, while ignoring the version field.
+   *
+   * @return The computed hash code without the version.
+   */
   public int hashWithoutVersion() {
     return Objects.hash(groupId, artifactId, classifier, packaging, scope);
   }
 
+  /**
+   * Checks if this Dependency object matches another Dependency without considering the version.
+   *
+   * This method compares the hash code of this Dependency with the hash code of another
+   * Dependency, excluding the version information. If the hash codes are equal, the two
+   * dependencies are considered a match without taking version into account.
+   *
+   * @param dependency The Dependency object to compare with this one.
+   * @return true if the dependencies match without considering the version, false otherwise.
+   */
   public boolean matchesWithoutConsideringVersion(final Dependency dependency) {
     return hashWithoutVersion() == dependency.hashWithoutVersion();
   }
--- a/plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/operator/Dependency.java
+++ b/plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/operator/Dependency.java
@@ -70,7 +70,7 @@ public class Dependency {
 
   @Override
   public int hashCode() {
-    return Objects.hash(groupId, artifactId, version, classifier, packaging, scope);
+    return Objects.hash(groupId, artifactId, classifier, packaging, scope);
   }
 
   public String getGroupId() {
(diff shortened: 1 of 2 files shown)


plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/operator/Dependency.java
```diff
@@ -70,7 +70,7 @@ public boolean equals(Object o) {
 
   @Override
   public int hashCode() {
```
nahsra, line 72: This feels wrong. I feel the contract for `hashCode` should be different if the `version` is different. If we have to make another method somewhere, like, `matchesWithoutConsideringVersion()`, then we should add that. But, I would look hard at that code.

plugins/codemodder-plugin-maven/src/test/java/io/codemodder/plugins/maven/operator/POMOperatorMultipomTest.java
```diff
@@ -12,49 +12,49 @@
...
 import org.junit.jupiter.api.Test;
 
 final class POMOperatorMultipomTest extends AbstractTestBase {
 
   @Test
   void testWithParentAndChildMissingPackaging() {
-    Assertions.assertThrows(
-        WrongDependencyTypeException.class,
-        () -> {
-          File parentResource = getResource("parent-and-child-parent-broken.xml");
```
nahsra, line 26: I would definitely prefer to `import static` these methods so they're easier to read.

plugins/codemodder-plugin-maven/src/test/java/io/codemodder/plugins/maven/operator/POMOperatorMultipomTest.java
```diff
@@ -12,49 +12,49 @@
...
+    Assertions.assertThatThrownBy(
+            () -> {
+              File parentResource = getResource("parent-and-child-parent-broken.xml");
 
-          List<POMDocument> parentPomFiles = Arrays.asList(POMDocumentFactory.load(parentResource));
+              List<POMDocument> parentPomFiles =
+                  Arrays.asList(POMDocumentFactory.load(parentResource));
```
nahsra, line 27: Does `spotless` run on this module?

plugins/codemodder-plugin-maven/src/test/java/io/codemodder/plugins/maven/operator/POMOperatorMultipomTest.java
```diff
@@ -12,49 +12,49 @@
...
-        });
+              gwt(
+                  "parent-and-child",
+                  parentPom.withDependency(Dependency.fromString("org.dom4j:dom4j:2.0.3")));
+            })
+        .isInstanceOf(WrongDependencyTypeException.class);
   }
 
   @Test
   void testWithParentAndChildWrongType() {
```
nahsra, line 40: These names don't really "tell the story" of the test they want to perform. Can we add some "human" descriptions (via JavaDoc) of the use case they are intended to support? Adding this kind of context to the code with comments is definitely part of bringing this module into the repo as it will make it much easier to maintain.

plugins/codemodder-plugin-maven/src/test/java/io/codemodder/plugins/maven/operator/POMOperatorMultipomTest.java
```diff
@@ -66,17 +66,15 @@ void testWithMultiplePomsBasicNoVersionProperty() throws Exception {
             .withParentPomFiles(Arrays.asList(POMDocumentFactory.load(parentPomFile)))
             .withUseProperties(false);
```
nahsra, line 68: I like this change. Nothing to stdout!

nahsra (changes requested):

plugins/codemodder-plugin-maven/src/test/java/io/codemodder/plugins/maven/operator/POMOperatorMultipomTest.java
carlosu7: I have run `./gradlew :plugins:codemodder-plugin-maven:spotlessApply` and it applies changes, does it mean yes???

nahsra: I mean when i just run `./gradlew spotlessApply` from the root, does it affect this module? If not, we should add it.

plugins/codemodder-plugin-maven/src/test/java/io/codemodder/plugins/maven/operator/POMOperatorMultipomTest.java
```diff
@@ -12,51 +15,66 @@
...
+        .isInstanceOf(WrongDependencyTypeException.class);
   }
 
+  /**
+   * Test case for a scenario with multiple POMs where the child POM does not have a version
+   * property, and properties are not used. This test validates dependency resolution and checks if
+   * the resulting POM document (pom-multiple-pom-basic-no-version-property-result.xml) is marked as
+   * dirty.
+   *
+   * @throws Exception If there is an exception during the test.
```
nahsra, line 76: We can just check not offer these `@throws` comments for tests

plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/operator/Dependency.java
```diff
@@ -73,6 +73,14 @@ public int hashCode() {
     return Objects.hash(groupId, artifactId, version, classifier, packaging, scope);
   }
```
nahsra, line 75: JavaDoc

plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/operator/Dependency.java
```diff
@@ -73,6 +73,14 @@ public int hashCode() {
...
 
+  public int hashWithoutVersion() {
+    return Objects.hash(groupId, artifactId, classifier, packaging, scope);
+  }
+
```
nahsra, line 79: JavaDoc

nahsra (approved):
I approved, but please fix the outstanding tasks

plugins/codemodder-plugin-maven/src/test/java/io/codemodder/plugins/maven/operator/POMOperatorMultipomTest.java
carlosu7: it does 👍

PR #216, opened 2023-10-27
Removing QueryByResolver

carlosu7:
Also removing test since it used QueryByResolver and QueryByEmbedder commands.

`QueryByResolver ` was used to resolve project dependencies using Apache Maven Aether by using Maven Embedder (EmbedderFacade) to build and resolve project models, then extract the resolved dependencies and returns them as a list.

At this point, we would only have one available dependency command in the code base: QueryByParsing which is a SAFE dependency command.


nahsra:
For each of these PRs, can you summarize what functionality we lose by removing this?

aldrinleal:
If you're removing these perhaps better review the dependencies since they pull a lot of those

carlosu7:
I'm removing all unused dependencies at the end in this PR: https://github.com/pixee/codemodder-java/pull/220

nahsra (approved):

PR #226, opened 2023-11-06
Refactor POM Operator

carlosu7:
Stop using static methods and move POM Operator related logic from MavenProvider to reduce public visibility of types

### DIFF
--- a/plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/MavenProvider.java
+++ b/plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/MavenProvider.java
@@ -193,28 +193,19 @@ public final class MavenProvider implements ProjectProvider {
               Collection<POMDocument> allPomFiles = modifiedProjectModel.allPomFiles();
               LOG.trace("Found " + allPomFiles.size() + " pom files -- " + allPomFiles);
               for (POMDocument aPomFile : allPomFiles) {
-                URI uri;
-                try {
-                  uri = aPomFile.getPomPath().toURI();
-                } catch (URISyntaxException ex) {
-                  ex.printStackTrace();
-                  throw new DependencyUpdateException("Failure parsing URL: " + aPomFile, ex);
-                }
+                final URI uri = getPomFileURI(aPomFile);
 
                 Path path = Path.of(uri);
 
                 if (aPomFile.getDirty()) {
-                  LOG.trace("POM file {} was dirty", path);
-                  try {
-                    CodeTFChangesetEntry entry = getChanges(projectDir, aPomFile, newDependencyGAV);
-                    pomModifier.modify(path, aPomFile.getResultPomBytes());
-                    LOG.trace("POM written!");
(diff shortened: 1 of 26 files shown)


plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/operator/FileUtils.java
```diff
@@ -0,0 +1,34 @@
+package io.codemodder.plugins.maven.operator;
```
carlosu7, line 1: this was created due to this sonarcloud hotspot issue https://sonarcloud.io/project/security_hotspots?id=pixee_codemodder-java&pullRequest=226&resolved=false&types=SECURITY_HOTSPOT&tab=code

plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/MavenProvider.java
```diff
@@ -163,128 +160,104 @@ private DependencyUpdateResult updateDependenciesInternal(
...
+      LOG.error("Failed to write pom", exc);
+      erroredFiles.add(path);
+    }
+  }
+
+  private URI getPomFileURI(final POMDocument aPomFile) {
+    try {
+      return aPomFile.getPomPath().toURI();
+    } catch (URISyntaxException ex) {
+      ex.printStackTrace();
```
nahsra, line 256:
```suggestion
LOG.error("Unexpected problem getting pom URI", e);
```

plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/operator/POMOperator.java
```diff
@@ -1,15 +1,89 @@
...
+   * Modifies and retrieves the ProjectModel with a new dependency.
+   *
+   * @param newDependencyGAV The new DependencyGAV to add to the POM.
+   * @return The modified ProjectModel, or null if the modification was unsuccessful.
+   * @throws XMLStreamException If an error occurs during XML stream processing.
+   * @throws URISyntaxException If there is an issue with the URI syntax.
+   * @throws IOException If an I/O error occurs.
+   * @throws DocumentException If an error occurs while parsing the document.
+   */
+  public ProjectModel modifyAndGetProjectModel(final DependencyGAV newDependencyGAV)
```
nahsra, line 44: The name here: `modifyAndGetProjectModel`

Can it be just `addDependency`? Isn't that all we're doing? We happen to return a `ProjectModel`, but it seems like it should be a simpler name.

nahsra (approved):
Approving, but consider name change

plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/operator/POMOperator.java
carlosu7: nvm... add dependency works

carlosu7:
sonarcloud is reporting to log the exception but it is been logged... merging this PR

PR #227, opened 2023-11-07
Refactor POMScanner

carlosu7:
Remove static methods

### DIFF
--- a/plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/operator/POMOperator.java
+++ b/plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/operator/POMOperator.java
@@ -2,10 +2,10 @@ package io.codemodder.plugins.maven.operator;
 
 import com.github.zafarkhaja.semver.Version;
 import io.codemodder.DependencyGAV;
+import java.io.File;
 import java.io.IOException;
 import java.net.URISyntaxException;
 import java.nio.file.Files;
-import java.nio.file.Path;
 import java.util.*;
 import java.util.stream.Collectors;
 import javax.xml.stream.XMLStreamException;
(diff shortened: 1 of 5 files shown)


plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/operator/POMOperator.java
```diff
@@ -14,22 +14,22 @@
 
 /** Facade for the POM Operator, providing methods for modifying and querying POM files. */
 public class POMOperator {
+  private final POMScanner pomScanner;
 
-  private final Path pomFile;
-  private final Path projectDir;
-
```
nahsra, line 20: We want to use `Path` as much as possible in `codemodder-java`. In fact, we only want to use `File` when its a 3rd party API that we don't control.

plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/operator/POMOperator.java
```diff
@@ -14,22 +14,22 @@
...
-  public POMOperator(final Path pomFile, final Path projectDir) {
-    this.pomFile = pomFile;
-    this.projectDir = projectDir;
+  public POMOperator(final File originalFile, final File topLevelDirectory) {
+    this.pomScanner = new POMScanner(originalFile, topLevelDirectory);
   }
```
nahsra, line 22: It feels like this method shouldn't exist. It's a hack to get past its own contract. A `POMOperator` without a `pom` and a `projectDir` doesn't make sense to me

plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/operator/POMOperator.java
carlosu7: I have a ticket created to replace File usage to Path

plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/operator/POMOperator.java
carlosu7: i don't understand... you mean `pomFile` and `projectDir` should be POMOperator class variables?

nahsra: I was talking about the `private POMOperator()` no-argument constructor. I think it should be deleted because it makes no sense. We also shouldn't have any testing-related methods here, like `forTesting()`. I think we should kill both and refactor whatever tests we need to make it work.

nahsra (approved):

PR #232, opened 2023-11-14
Command chain to insert only a dependency

carlosu7:
Adding new method `insertDependency` to POMOperator to only insert a dependency if it is not present in the pom


plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/operator/FormatCommand.java
```diff
@@ -502,9 +502,12 @@ private byte[] serializePomFile(POMDocument pom) throws XMLStreamException {
       Integer key = entry.getKey();
       MatchData match = entry.getValue();
 
-      MatchData nextMatch = elementsToReplace.remove(0);
+      if (elementsToReplace != null && elementsToReplace.size() > 0) {
+        MatchData nextMatch = elementsToReplace.remove(0);
 
-      xmlRepresentation = replaceRange(xmlRepresentation, match.getRange(), nextMatch.getContent());
+        xmlRepresentation =
```
carlosu7, line 508: this change was to handle a index out of bounds exception

plugins/codemodder-plugin-maven/src/test/java/io/codemodder/plugins/maven/operator/POMOperatorTest.java
```diff
@@ -519,4 +519,68 @@ void testCaseWithEmptyElementFromCustomer() throws Exception {
...
+        new Dependency("org.dom4j", "dom4j", "2.0.2", null, null, null);
+
+    ProjectModel context =
+        performInsert(
+            "pom-case-3-insert-fails",
+            ProjectModelFactory.load(POMOperatorTest.class.getResource("pom-case-3.xml"))
+                .withDependency(dependencyToUpgrade)
+                .withSkipIfNewer(true));
+
+    Assert.assertTrue("Original POM File is Dirty", !context.getPomFile().getDirty());
```
nahsra, line 535: This logic is weird, can we write it so it's not a double negative?
```suggestion
    Assert.assertFalse("Original POM File is Dirty", context.getPomFile().getDirty());
```

plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/operator/FormatCommand.java
nahsra: Is there a regression test for this?

nahsra (changes requested):

plugins/codemodder-plugin-maven/src/test/java/io/codemodder/plugins/maven/operator/POMOperatorTest.java
carlosu7: it is weird... i got confused at some point while i was having some issues

plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/operator/FormatCommand.java
carlosu7: i removed this condition to validate where the exception was thrown and write a test but it is not being thrown... It happened at the beginning of my research but it got fixed at some point during development process

nahsra (approved):

PR #235, opened 2023-11-16
Release integration test framework - generate images during check action execution

arturoespinoza15:
**Overview:**
Test framework was meant to run the integration tests during CI/CD pipelines but currently the tests can only run locally by following the README. In order to release the test framework we need to build a infrastructure that allow us to upload the required images to be used by the framework.

Cases:

- Since the codemodder repo is public, all the people should be allowed to run the integration tests, so for the case when the tests are run locally, test projects images and codemodder base image should have a latest image uploaded to be used. Images should be public in order to not required a specific config to be pulled. Latests images can be uploaded an versioned each time a PR is merged and those images can be used when the tests are run locally, so the images that are using contains the most recent changes in `main` .
- We want to run the integration tests in each pull request in order to verify PR changes does not affect codemods functionality. We need to generate dynamically the images with the PR changes but we don’t want upload the images to the registry in each commit because we don’t need a lot of images that only will be used once and the versioning would be difficult. In this case we want to build the images and only have them available in the action context.

**What changed:**
- this PR solves the second case commented above, in each PR we want to run the integration tests to verify that codemods are no broken so we need to generate the images from the changes in the PR
- additonal steps were added to `checks.yml` for the images build
- gradle task is added to generate a list of the test projects names for the images build, this way each time a new integration test is added, test project image used for that test will be generated automatically
- Some changes were applied to the codemodder base image dockerfile in order to improve the build time

**How to test:**
- existing integration tests are enabled so the images built in the action context are used in the integration tests executions, tests should pass


core-codemods/src/test/java/io/codemodder/codemods/integration/baseimage/Dockerfile
```diff
@@ -19,16 +16,11 @@ RUN wget -q "https://services.gradle.org/distributions/gradle-${GRADLE_VERSION}-
...
-RUN python --version
-RUN gradle --version
-
-# Install Semgrep using pip
-RUN pip install semgrep
+RUN ln -s /usr/bin/python3 /usr/bin/python && \
+    java -version && \
+    python --version && \
+    gradle --version && \
+    pip install semgrep
```
nahsra, line 24: I think we need to pin semgrep's version as we do above:

`python3 -m pip install semgrep==1.15.0`

Not sure how this isn't causing issues.

.github/workflows/checks.yml
```diff
@@ -4,22 +4,79 @@ on:
...
+      - name: Build codemodder base image
+        uses: docker/build-push-action@0565240e2d4ab88bba5387d719585280857ece09 # v5.0.0
+        with:
+          context: .
+          file: ${{ env.TEST_FRAMEWORK_PATH }}/baseimage/Dockerfile
+          tags: codemodder/codemodder-base:latest
+
+      - name: Setup test project names
+        run: |
+          CODEMOD_IDS=$(./gradlew core-codemods:getTestProjectNames -q)
```
nahsra, line 54: Can you explain this build step? What are we injecting the codemod IDs into, and why? Maybe we need more docs here.

core-codemods/build.gradle.kts
```diff
@@ -67,3 +67,32 @@ tasks.named("check") {
     @Suppress("UnstableApiUsage")
     dependsOn(testing.suites.named(integrationTestSuiteName))
 }
+
```
nahsra, line 70: Let me make sure I understand what I'm looking at here. The `getTestProjectNames` function runs and creates a list of all the codemod IDs. This is now a dependency of the compile task, so if that list changes (e.g., we introduce a new one), then new images will be built. Is that right?

What happens if I just alter a test file in the integration test app itself? Will Gradle know to re-build the image during the pipeline, and deploy a new one on release?

.github/workflows/checks.yml
```diff
@@ -4,22 +4,79 @@ on:
...
+  TEST_FRAMEWORK_PATH: core-codemods/src/test/java/io/codemodder/codemods/integration
+
 jobs:
   check:
     name: Check
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
-      - uses: actions/setup-python@v4
+      - name: Clone repository
```
nahsra, line 15: Why are we pinning the commit rather than the version, is that the new best practice?

nahsra (dismissed):
I think this is probably fine, but I'm not that strong with Gradle so may need some help reviewing + more comments.

nahsra:
Also, if this is ready, let's hit "Ready for review".

.github/workflows/checks.yml
arturoespinoza15: Github suggest that using the SHA instead the version is more reliable because we are pinned to a stable version https://docs.github.com/en/actions/learn-github-actions/finding-and-customizing-actions#using-shas

core-codemods/build.gradle.kts
arturoespinoza15: the task extracts the codemod IDs from the existing integration tests, remember that each existing test app has the id of the comedmod that is testing  as name, so yes the output will be a list with the codemod IDs and will be used to build the respective image for each integration test. If I create a new integration test for a codemod, its test project must be created also and the pipeline will automatically build a new test project image for the new test.

for the question about what happens when a test app is updated, since each test app image is using the codemodder-base image as base, in each change that affects the codemodder-java(including changes over the test apps) then the codemodder-base image will be re-built and consequently the test projects images will be re-built also

.github/workflows/checks.yml
arturoespinoza15: as I commented, the task output is the list of the project tests names that will be iterated in the `Build test projects images` step in order to build the respective images, maybe the variable name is confused, I will rename it

core-codemods/src/test/java/io/codemodder/codemods/integration/baseimage/Dockerfile
arturoespinoza15: do we want that specific version?

nahsra: Yes, installing later versions breaks stuff.

arturoespinoza15:
> Approved, but we need to pin Semgrep version in order to not break some of our existing codemods.

yes, the version was pinned

PR #236, opened 2023-11-16
:arrow_up: upgrade java-security-toolkit java to 1.1.1

ryandens:


framework/codemodder-base/src/main/java/io/codemodder/DependencyGAV.java
```diff
@@ -160,7 +160,7 @@ static DependencyGAV createDefault(
         group, artifact, version, justification, license, repositoryUrl, noTransitiveDependencies);
   }
 
-  String JAVA_SECURITY_TOOLKIT_VERSION = "1.0.7";
+  String JAVA_SECURITY_TOOLKIT_VERSION = "1.1.1";
```
ryandens, line 163: we could do some fancy stuff here to template this into the code at build time, but not worthwhile for right now IMO. We should just be vigilant about upgrading

nahsra (approved):

PR #237, opened 2023-11-16
Release integration test framework - generate and upload official images 

arturoespinoza15:
**Overview:**
Test framework was meant to run the integration tests during CI/CD pipelines but currently the tests can only run locally by following the README. In order to release the test framework we need to build a infrastructure that allow us to upload the required images to be used by the framework.

Cases:

- Since the codemodder repo is public, all the people should be allowed to run the integration tests, so for the case when the tests are run locally, test projects images and codemodder base image should have a latest image uploaded to be used. Images should be public in order to not required a specific config to be pulled. Latests images can be uploaded an versioned each time a PR is merged and those images can be used when the tests are run locally, so the images that are using contains the most recent changes in `main` .
- We want to run the integration tests in each pull request in order to verify PR changes does not affect codemods functionality. We need to generate dynamically the images with the PR changes but we don’t want upload the images to the registry in each commit because we don’t need a lot of images that only will be used once and the versioning would be difficult. In this case we want to build the images and only have them available in the action context.

**What changed:**
- this PR solves the first case commented above, github action is added to generate the images required for the framework 
with the latest changes and push them to the public registry as the official and latest images. These images will be pulled and used during the integration test execution locally


.github/workflows/release-integration-test-framework-images.yml
```diff
@@ -0,0 +1,83 @@
...
+#     tar file generation, tar file contains codemodder executable file to be used in the codemodder-base image
+      - name: Generate distribution
+        uses: gradle/gradle-build-action@842c587ad8aa4c68eeba24c396e15af4c2e9f30a # v2.9.0
+        with:
+          arguments: core-codemods:distTar
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@343f7c4344506bcbf9b4de18042ae17996df046d # v3.0.0
+        with:
+          username: ${{ vars.DOCKERHUB_USERNAME }}
```
arturoespinoza15, line 49: before merge this PR, I would need you to add these values at the repo level. I have talked with David to create a user in the codemodder organization in dockerhub. Username should be stored as a repository variable and a generated access token should be stored as repository secret

nahsra: Can you work with @dhafley directly on this? He has the access to add repository variables and secrets here.

arturoespinoza15: sure, I already passed him the names and values for the secret and variable I need, just waiting for that

PR #238, opened 2023-11-16
:sparkles: Apache 2 Community Codemods

ryandens:
- :truck: Rename examples project to community-codemods
- :rocket: publish community codemods with apache license


nahsra (approved):

gradle/build-plugins/src/main/kotlin/io.codemodder.maven-publish.gradle.kts
```diff
@@ -42,10 +42,12 @@ publishing {
                 }
             }
             pom {
-                licenses {
-                    license {
-                        name.set("GNU AFFERO GENERAL PUBLIC LICENSE, Version 3 (AGPL-3.0)")
-                        url.set("https://opensource.org/licenses/AGPL-3.0")
+                if (!pluginManager.hasPlugin("com.netflix.nebula.maven-apache-license")) {
```
ryandens, line 45: we could also restructure this a bit to have a `maven-publish-base` plugin which does everything except configure the pom license. Then, we have two high-level plugins `maven-publish` which configures it with AGPL and `maven-publish-apache` which configures it with apache 2

ryandens:
Sonar doesn't like the duplication that exists mainly as a result of the boilerplate associated with declaring a codemodder. I'm fine to ignore in this context - but i think it's a sign that we have API improvements to make here

PR #239, opened 2023-11-17
Update CodemodderOpenAIKeys.java  This provides additional informatio…

Yesahk:
…n in case the key is not available.

### DIFF
--- a/framework/codemodder-testutils-llm/src/main/java/io/codemodder/testutils/llm/CodemodderOpenAIKeys.java
+++ b/framework/codemodder-testutils-llm/src/main/java/io/codemodder/testutils/llm/CodemodderOpenAIKeys.java
@@ -13,6 +13,13 @@ public final class CodemodderOpenAIKeys {
    */
   public static boolean isAvailable() {
     String key = System.getenv("CODEMODDER_OPENAI_API_KEY");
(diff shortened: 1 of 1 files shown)


framework/codemodder-testutils-llm/src/main/java/io/codemodder/testutils/llm/CodemodderOpenAIKeys.java
```diff
@@ -13,6 +13,13 @@ public final class CodemodderOpenAIKeys {
    */
   public static boolean isAvailable() {
     String key = System.getenv("CODEMODDER_OPENAI_API_KEY");
```
nahsra, line 15: What type of behavior do you think would be best for codemod authors? I'm only working on OpenAI-based codemods sometimes, so having a loud error (or throwing an exception) when there's not a key available feels heavy handed.

Yesahk:
One of the behaviors that I think would be best for codemod authors are
as follows,
*Return Optional or Result Types:* Instead of returning a boolean, you
could return an Optional<String> or a custom result type that encapsulates
the presence or absence of the API key. This allows the calling code to
easily check for the presence of the key and obtain its value if available.

PR #241, opened 2023-11-21
New method to only update the existing dependencies of a POM file

carlosu7:
New method  `POMOperator::update` to update only the existing dependencies of a POM file.


plugins/codemodder-plugin-maven/src/test/java/io/codemodder/plugins/maven/operator/POMOperatorTest.java
```diff
@@ -690,4 +690,140 @@ void insert_should_perform_gracefully() throws Exception {
...
+    ProjectModel context =
+        performAndAssertUpdatePomOperation("case-3-no-dependency-update", projectModelFactory);
+
+    Diff diff =
+        getXmlDifferences(
+            context.getPomFile().getPomDocument(), context.getPomFile().getResultPom());
+
+    // "Original POM File is Dirty"
+    assertThat(context.getPomFile().getDirty()).isFalse();
+    ;
```
nahsra, line 798: Dangling semi

plugins/codemodder-plugin-maven/src/test/java/io/codemodder/plugins/maven/operator/POMOperatorTest.java
```diff
@@ -690,4 +690,140 @@ void insert_should_perform_gracefully() throws Exception {
     assertThat(resolvedDeps.size()).isOne();
     assertThat(resolvedDeps.get(0)).isEqualTo(dependencyToUpgrade);
   }
+
```
nahsra, line 693: Imagine we have two codemods. 

The first attempts to install `gson:1.2` and the next attempts to install `gson:1.1`, what will the result be?

Maybe I am questioning whether we should allow 2 different commands for insert and update, differently. The whole point of bringing `pom-operator` in here is to craft it towards codemodder's needs.

And it seems like the use case from codemodder's side is _always_ "insert new or update what's there to match the newer version I propose" (like `UPSERT` in SQL). Is that not true? Am I missing something?

plugins/codemodder-plugin-maven/src/test/java/io/codemodder/plugins/maven/operator/POMOperatorTest.java
```diff
@@ -690,4 +690,140 @@ void insert_should_perform_gracefully() throws Exception {
...
+        getXmlDifferences(
+            context.getPomFile().getPomDocument(), context.getPomFile().getResultPom());
+
+    // "Original POM File is Dirty"
+    assertThat(context.getPomFile().getDirty()).isTrue();
+    ;
+
+    // "Document has differences"
+    assertThat(diff.hasDifferences()).isTrue();
+    ;
```
nahsra, line 716: dangling

plugins/codemodder-plugin-maven/src/test/java/io/codemodder/plugins/maven/operator/POMOperatorTest.java
```diff
@@ -690,4 +690,140 @@ void insert_should_perform_gracefully() throws Exception {
...
+    }
+
+    assertThat(differenceList.size()).isOne();
+
+    // "Document changes a single version"
+    assertThat(differenceList.get(0).toString())
+        .startsWith("Expected text value '0.0.1-SNAPSHOT' but was '1.0.0'");
+
+    // "Document changes a property called 'sample.version'"
+    assertThat(differenceList.get(0).getComparison().getTestDetails().getXPath())
```
nahsra, line 775: This is an interesting way to record the changes, cool!

nahsra (changes requested):
So, this change makes updates

plugins/codemodder-plugin-maven/src/test/java/io/codemodder/plugins/maven/operator/POMOperatorTest.java
carlosu7: @nahsra 

> The first attempts to install gson:1.2 and the next attempts to install gson:1.1, what will the result be?

it depends on how the project model is built by the ProjectModelFactory, there's a property called "withSkipIfNewer" that will check if we want to skip the version update 

so if we set it to true, in your example gson:1.2 will be the final version

carlosu7: > Maybe I am questioning whether we should allow 2 different commands for insert and update, differently. The whole point of bringing pom-operator in here is to craft it towards codemodder's needs.
> 
> And it seems like the use case from codemodder's side is always "insert new or update what's there to match the newer version I propose" (like UPSERT in SQL). Is that not true? Am I missing something?

@nahsra yes, right now, codemodder just sents a lists of dependencies and POMOperator executes `modify` method to "insert new or update dependency"... in the case of updating the version, right now it always sets/leaves the newest


i'm not sure if that answers you're question

nahsra (approved):

PR #245, opened 2023-11-28
Create dist action for testing

nahsra:
We need a new release script just for testing the distribution package to GitHub Releases. Although this first draft (almost certainly) won't work, once it's in `main` we can iterate on it from the Actions tab (I think).

### DIFF
--- a/.github/workflows/release_without_maven.yml
+++ b/.github/workflows/release_without_maven.yml
@@ -1,0 +1,67 @@
+name: Release to GitHub
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: a stable version to set in version.txt before publishing
+
+jobs:
+  release:
+    name: Build & Release
+    permissions:
+      contents: write
+      id-token: write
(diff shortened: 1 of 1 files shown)


.github/workflows/release_without_maven.yml
```diff
@@ -0,0 +1,67 @@
...
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+
+      - name: Install Semgrep
+        run: python3 -m pip install semgrep
```
drdavella, line 31: Is this actually necessary for the build step? We can definitely include this dependency in the CLI package.

drdavella (approved):

.github/workflows/release_without_maven.yml
nahsra: Yes, because our build runs tests, which tests codemods, which require Semgrep to execute successfully.

PR #247, opened 2023-11-28
New Sonarcloud codemod to substitute `String#replaceAll()`

carlosu7:
This change replaces the usage of `String#replaceAll()` to `String#replace()` when the first argument is not a regular expression. It does exactly the same thing as `String#replaceAll()` without the performance drawback of the regex.


core-codemods/src/main/resources/io/codemodder/codemods/ReplaceCodemod/description.md
```diff
@@ -0,0 +1,10 @@
+This change replaces the usage of `String::replaceAll` to `String::replace` when the first argument is not a regular expression. It does exactly the same thing as String::replaceAll` without the performance drawback of the regex.
+
+Our changes look something like this:
```
nahsra, line 3: Are these originally written? We can't take text from anyone else.

core-codemods/src/main/resources/io/codemodder/codemods/SubstituteReplaceAllCodemod/description.md
```diff
@@ -0,0 +1,10 @@
+This change substitutes the usage of `String#replaceAll()` to `String#replace()`.
```
nahsra, line 1: Maybe ask ChatGPT for a concise explanation of the value here?

nahsra: It should match the level of detail in the other `description.md` files.

core-codemods/src/main/resources/io/codemodder/codemods/SubstituteReplaceAllCodemod/description.md
nahsra, line 1: God, ChatGPT is so verbose. Here's my version:

```
This change replaces `String#replaceAll()` with `String#replace()` to enhance performance and avoid confusion. 

The `String#replaceAll()` call takes a regular expression for the first argument, which is then compiled and used to replace string subsections. However, the argument being passed to it doesn't actually appear to be a regular expression. Therefore, the `replace()` [API](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#replace-java.lang.CharSequence-java.lang.CharSequence-) appears to be a better fit.

Our changes look something like this:
```

nahsra (changes requested):

core-codemods/src/main/resources/io/codemodder/codemods/SubstituteReplaceAllCodemod/report.json
```diff
@@ -0,0 +1,7 @@
+{
```
nahsra, line 1: I think these should be:

```
Fixed inefficient usage of `String#replaceAll()`
```

nahsra (approved):
Approving, but please make the given change!

PR #249, opened 2023-11-29
Update release name and documentation

nahsra:
* Changes release and tag names to be the simple semantic version.
* Adds documentation to the `README` on how to use it as a CLI.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -51,8 +51,8 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
-          tag_name: ${{ inputs.version }}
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -51,8 +51,8 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
-          tag_name: ${{ inputs.version }}
-          release_name: Release ${{ inputs.version }}
+          tag_name: v${{ inputs.version }}
```
drdavella, line 54: Generally I like to avoid `v` in the tag name and just have the bare version itself but I don't feel too strongly about it.

drdavella (approved):

PR #253, opened 2023-12-04
Sonar codemod to remove redundant variable creation expression

carlosu7:
Remove redundant variable creation expression when it is only returned or thrown

[Given examples of the file](https://sonarcloud.io/project/issues?fileUuids=AYvtrjqJLCzGLicz7AcH&resolved=false&types=CODE_SMELL&id=nahsra_WebGoat_10_23):

[Immediately return this expression instead of assigning it to the temporary variable "localeResolver".](https://sonarcloud.io/project/issues?open=AYvtrj5QLCzGLicz7AqQ&id=nahsra_WebGoat_10_23)
[Immediately throw this expression instead of assigning it to the temporary variable "myException".](https://sonarcloud.io/project/issues?open=AYw20oU9yE-xQ1mlDtlZ&id=nahsra_WebGoat_10_23)

Sonar rule reference: https://rules.sonarsource.com/java/RSPEC-1488/


core-codemods/src/main/resources/io/codemodder/codemods/RemoveRedundantVariableCreationCodemod/description.md
```diff
@@ -0,0 +1,19 @@
+TODO
```
carlosu7, line 1: @nahsra you're better at this than me u.u

core-codemods/src/main/java/io/codemodder/codemods/RemoveRedundantVariableCreationCodemod.java
```diff
@@ -0,0 +1,65 @@
...
+  @Override
+  public boolean onIssueFound(
+      final CodemodInvocationContext context,
+      final CompilationUnit cu,
+      final ObjectCreationExpr objectCreationExpr,
+      final Issue issue) {
+
+    // Get full block statement
+    final Optional<BlockStmt> blockStmtOpt = objectCreationExpr.findAncestor(BlockStmt.class);
+    if (blockStmtOpt.isPresent()) {
```
nahsra, line 37: This logic doesn't feel right to me. This will always return `true`, even in a situation where no changes were made.

core-codemods/src/main/java/io/codemodder/codemods/RemoveRedundantVariableCreationCodemod.java
```diff
@@ -0,0 +1,65 @@
...
+      final BlockStmt blockStmt = blockStmtOpt.get();
+
+      // Retrieve return/throw statement to update its expression to given objectCreationExpr param
+      final Optional<Statement> lastStmtOpt =
```
nahsra, line 41: This logic feels shakey too. If we know for certain the one we're looking for is always the last statement, why don' we call `blockStmt.getStatements().get(size)` to get it? This filtering seems like a weird way to find what we know we can find by looking at the end, right?

core-codemods/src/main/java/io/codemodder/codemods/RemoveRedundantVariableCreationCodemod.java
```diff
@@ -0,0 +1,65 @@
...
+          blockStmt.getStatements().stream()
+              .filter(stmt -> stmt instanceof ReturnStmt || stmt instanceof ThrowStmt)
+              .findFirst();
+
+      lastStmtOpt.ifPresent(
+          lastStmt -> {
+            final Expression expression = objectCreationExpr.clone();
```
nahsra, line 48: Hmm, why `clone()`? I've never had to invoke this, so I'm curious. I wonder if this will retain the concrete syntax (spacing) of the original node.

core-codemods/src/main/resources/io/codemodder/codemods/RemoveRedundantVariableCreationCodemod/description.md
nahsra:
```
This change removes intermediate variables who are only created to be thrown or returned in the next statement. This makes the code more readable, which makes reviewing the code for issues easier.
```

nahsra (changes requested):

core-codemods/src/main/java/io/codemodder/codemods/RemoveRedundantVariableCreationCodemod.java
carlosu7: since it is sorted, yes, i didn't considered that

core-codemods/src/main/java/io/codemodder/codemods/RemoveRedundantVariableCreationCodemod.java
carlosu7: @nahsra I tried doing it this way:

```
if (lastStmt instanceof ReturnStmt returnStmt) {
              returnStmt.setExpression(expression);
            }
            if (lastStmt instanceof ThrowStmt throwStmt) {
              throwStmt.setExpression(expression);
            }
```

but my test fails 😟

carlosu7: updated xd

core-codemods/src/main/java/io/codemodder/codemods/RemoveRedundantVariableCreationCodemod.java
```diff
@@ -0,0 +1,65 @@
...
+      // Retrieve the redundant variable declaration expression that will be removed
+      final Optional<ExpressionStmt> exprStmtOpt =
+          objectCreationExpr.findAncestor(ExpressionStmt.class);
+
+      if (lastStmtOpt.isPresent() && exprStmtOpt.isPresent()) {
+        final Statement lastStmt = lastStmtOpt.get();
+        if (lastStmt instanceof ReturnStmt returnStmt) {
+          returnStmt.setExpression(objectCreationExpr);
+        }
+        if (lastStmt instanceof ThrowStmt throwStmt) {
```
nahsra, line 52: Should this be `else if`? I feel like we should have an `else` block here too, in case its something unexpected (and which we shouldn't call the `remove()`:
```
if(isReturn) {
 ...
} else if(isThrow) {
 ...
} else {
 // something weird happened, don't remove
}
```

Am I crazy?

nahsra (changes requested):

core-codemods/src/main/java/io/codemodder/codemods/RemoveRedundantVariableCreationCodemod.java
carlosu7: nop, something weird could happen

nahsra (approved):

PR #256, opened 2023-12-06
Sonar codemod to Add a private constructor to hide the implicit public one.

carlosu7:
Utility classes should not have public constructors

### DIFF
--- a/core-codemods/src/main/java/io/codemodder/codemods/AvoidImplicitPublicConstructorCodemod.java
+++ b/core-codemods/src/main/java/io/codemodder/codemods/AvoidImplicitPublicConstructorCodemod.java
@@ -1,0 +1,35 @@
+package io.codemodder.codemods;
+
+import com.github.javaparser.ast.CompilationUnit;
+import com.github.javaparser.ast.Node;
+import io.codemodder.*;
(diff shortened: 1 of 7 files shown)


core-codemods/src/main/java/io/codemodder/codemods/AvoidImplicitPublicConstructorCodemod.java
```diff
@@ -0,0 +1,60 @@
...
+import io.codemodder.providers.sonar.RuleIssues;
+import io.codemodder.providers.sonar.SonarPluginJavaParserChanger;
+import io.codemodder.providers.sonar.api.Issue;
+import java.util.Optional;
+import javax.inject.Inject;
+
+/** A codemod for setting a private constructor to hide implicit public constructor (Sonar) */
+@Codemod(
+    id = "sonar:java/avoid-implicit-public-constructor-s1118",
+    reviewGuidance = ReviewGuidance.MERGE_WITHOUT_REVIEW,
```
nahsra, line 22: This one should definitely be `MERGE_AFTER_REVIEW`.

nahsra (changes requested):

nahsra (approved):

PR #259, opened 2023-12-08
Avoid NPE when finding for file

carlosu7:
Reported NPE is caused by the project directory's parent `projectDir.getParent()` in `!Files.isSameFile(projectDir.getParent(), parent)`

```
io.codemodder.plugins.maven.MavenProvider$DependencyUpdateException: Failure when updating dependencies
	at io.codemodder.plugins.maven.MavenProvider.updateDependencies(MavenProvider.java:125)
	at io.codemodder.DefaultCodemodExecutor.addPackages(DefaultCodemodExecutor.java:300)
	at io.codemodder.DefaultCodemodExecutor.updateFiles(DefaultCodemodExecutor.java:218)
	at io.codemodder.DefaultCodemodExecutor.lambda$execute$0(DefaultCodemodExecutor.java:149)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:840)
Caused by: java.lang.NullPointerException: Cannot invoke "java.nio.file.Path.getFileSystem()" because "path" is null
	at java.base/java.nio.file.Files.provider(Files.java:105)
	at java.base/java.nio.file.Files.isSameFile(Files.java:1541)
	at io.codemodder.plugins.maven.MavenProvider$DefaultPomFileFinder.findForFile(MavenProvider.java:102)
	at io.codemodder.plugins.maven.DefaultPOMDependencyUpdater.isEmptyPomFile(DefaultPOMDependencyUpdater.java:128)
	at io.codemodder.plugins.maven.DefaultPOMDependencyUpdater.execute(DefaultPOMDependencyUpdater.java:75)
	at io.codemodder.plugins.maven.MavenProvider.updateDependencies(MavenProvider.java:121)
	... 8 more
```

### DIFF
--- a/plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/MavenProvider.java
+++ b/plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/MavenProvider.java
@@ -103,7 +103,7 @@ public final class MavenProvider implements ProjectProvider {
 
         Path parent = file.getParent();
 
-        while (parent != null && isDifferentFile(projectDir, parent)) {
+        while (parent != null && isProjectDirParentNullOrDifferentToParent(projectDir, parent)) {
           final Path pomPath = parent.resolve("pom.xml");
           if (Files.exists(pomPath)) {
             return Optional.of(pomPath);
@@ -115,7 +115,8 @@ public final class MavenProvider implements ProjectProvider {
       return Optional.empty();
     }
 
-    private boolean isDifferentFile(final Path projectDir, final Path parent) throws IOException {
+    private boolean isProjectDirParentNullOrDifferentToParent(
+        final Path projectDir, final Path parent) throws IOException {
       final Path projectDirParent = projectDir.getParent();
       return projectDirParent == null || !Files.isSameFile(projectDirParent, parent);
     }
--- a/plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/MavenProvider.java
+++ b/plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/MavenProvider.java
@@ -99,10 +99,12 @@ public final class MavenProvider implements ProjectProvider {
     @Override
     public Optional<Path> findForFile(final Path projectDir, final Path file) throws IOException {
 
-      if (projectDir != null && file != null && projectDir.getParent() != null) {
+      if (projectDir != null && file != null) {
+
         Path parent = file.getParent();
-        while (parent != null && !Files.isSameFile(projectDir.getParent(), parent)) {
-          Path pomPath = parent.resolve("pom.xml");
+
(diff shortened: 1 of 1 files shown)


plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/MavenProvider.java
```diff
@@ -98,16 +98,27 @@ public MavenProvider() {
   static class DefaultPomFileFinder implements PomFileFinder {
     @Override
     public Optional<Path> findForFile(final Path projectDir, final Path file) throws IOException {
-      Path parent = file.getParent();
-      while (parent != null && !Files.isSameFile(projectDir.getParent(), parent)) {
-        Path pomPath = parent.resolve("pom.xml");
-        if (Files.exists(pomPath)) {
-          return Optional.of(pomPath);
+
+      if (projectDir != null && file != null) {
```
carlosu7, line 102: add NPE assertion for given params

plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/MavenProvider.java
```diff
@@ -98,16 +98,27 @@ public MavenProvider() {
...
+
+        Path parent = file.getParent();
+
+        while (parent != null && isDifferentFile(projectDir, parent)) {
```
carlosu7, line 106: previous second condition : `!Files.isSameFile(projectDir.getParent(), parent)`

was checking if `projectDir.getParent()` and `parent` were different 

so I updated it to : `projectDirParent == null || !Files.isSameFile(projectDirParent, parent)`

because if projectDirParent is null, it will mean that projectDirParent and parent are different and there's no need to use` !Files.isSameFile` to check if those paths are different

carlosu7:
Asked chatgpt about scenarios where getParent can be null

>  a few scenarios where getParent() might return null:
> 
> Root Directory: If the provided projectDir represents the root directory (e.g., "C:" on Windows or "/" on Unix-based systems), its parent directory doesn't exist, so getParent() returns null.
> 
> Path Represents a File: If projectDir represents a file rather than a directory, its parent directory might not exist, causing getParent() to return null.
> 
> Path Doesn't Have a Parent: For certain paths without a parent (like a relative path that represents only a file name without any directory structure), getParent() will return null.

it would be valuable to have more info about how it was run codemodder for the liferay-portal repo to see how the path location is being sent @nahsra

nahsra:
Can we move this to draft while we wait to hear more? I think open PRs should represent work that shortly will move to `main`.

PR #261, opened 2023-12-13
Sonar codemod for removing unused local variable

carlosu7:
Codemod to remove unused local variables which expression is a variable (`NameExpr`) or just a Literal expression like a single boolean, char, double, integer, long, null, string or a text block string. We are not considering create object expression, method call expressions, condition expressions, etc. because all of them have an expression node and that expression node could result in a method call expression where a process could be performed and deleting it could result on some unexpected behaviors.

Examples of unused `LiteralExpr` : 
* String msg = "my message"
* char myChar = 'c';
* int ten = 10;
* etc

sonar cloud report example: https://sonarcloud.io/project/issues?fileUuids=AYvtrjqJLCzGLicz7Abq&resolved=false&types=CODE_SMELL&id=nahsra_WebGoat_10_23

reference: https://rules.sonarsource.com/java/RSPEC-1481/

### DIFF
--- a/core-codemods/src/main/java/io/codemodder/codemods/RemoveUnusedLocalVariableCodemod.java
+++ b/core-codemods/src/main/java/io/codemodder/codemods/RemoveUnusedLocalVariableCodemod.java
@@ -1,0 +1,36 @@
+package io.codemodder.codemods;
+
+import com.github.javaparser.ast.CompilationUnit;
+import com.github.javaparser.ast.Node;
+import com.github.javaparser.ast.expr.SimpleName;
+import io.codemodder.*;
+import io.codemodder.providers.sonar.ProvidedSonarScan;
+import io.codemodder.providers.sonar.RuleIssues;
+import io.codemodder.providers.sonar.SonarPluginJavaParserChanger;
+import io.codemodder.providers.sonar.api.Issue;
+
+import javax.inject.Inject;
+
+/**   */
+@Codemod(
+    id = "sonar:java/remove-unused-local-variable-s1481",
+    reviewGuidance = ReviewGuidance.MERGE_WITHOUT_REVIEW,
+    executionPriority = CodemodExecutionPriority.HIGH)
+public final class RemoveUnusedLocalVariableCodemod extends SonarPluginJavaParserChanger<Node> {
+
+  @Inject
+  public RemoveUnusedLocalVariableCodemod(
+      @ProvidedSonarScan(ruleId = "java:S1481") final RuleIssues issues) {
+    super(issues, Node.class, RegionNodeMatcher.MATCHES_START);
+  }
+
+  @Override
+  public boolean onIssueFound(
+      final CodemodInvocationContext context,
+      final CompilationUnit cu,
+      final Node name,
+      final Issue issue) {
+
+    return true;
+  }
+}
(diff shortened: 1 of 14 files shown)


core-codemods/src/main/java/io/codemodder/codemods/RemoveUnusedLocalVariableCodemod.java
```diff
@@ -0,0 +1,63 @@
...
+  public boolean onIssueFound(
+      final CodemodInvocationContext context,
+      final CompilationUnit cu,
+      final VariableDeclarator variableDeclarator,
+      final Issue issue) {
+
+    final Optional<Expression> initializer = variableDeclarator.getInitializer();
+
+    if (initializer.isPresent()) {
+      final Expression initializerExpr = initializer.get();
```
nahsra, line 46: Shouldn't the `return true` be line 54 or 55? In all other cases, nothing is removed.

core-codemods/src/main/resources/io/codemodder/codemods/RemoveUnusedLocalVariableCodemod/description.md
```diff
@@ -0,0 +1,11 @@
+TODO
```
nahsra, line 1:
```
This change removes unused variables. Unused variables make the code harder to read, which will lead to confusion and bugs. We only remove variables that have no state-changing effects.
```

core-codemods/src/main/java/io/codemodder/codemods/RemoveUnusedLocalVariableCodemod.java
```diff
@@ -0,0 +1,63 @@
...
+
+      if (initializerExpr instanceof LiteralExpr || initializerExpr instanceof NameExpr) {
+        final Optional<Node> variableDeclarationExprOptional = variableDeclarator.getParentNode();
+
+        if (variableDeclarationExprOptional.isPresent()) {
```
nahsra, line 51: Does this handle the case where:

```
int a =1, b=2;
```

Where only `a` is unused?

core-codemods/src/main/java/io/codemodder/codemods/RemoveUnusedLocalVariableCodemod.java
```diff
@@ -0,0 +1,63 @@
...
+@Codemod(
+    id = "sonar:java/remove-unused-local-variable-s1481",
+    reviewGuidance = ReviewGuidance.MERGE_WITHOUT_REVIEW,
+    executionPriority = CodemodExecutionPriority.HIGH)
+public final class RemoveUnusedLocalVariableCodemod
+    extends SonarPluginJavaParserChanger<VariableDeclarator> {
+
+  @Inject
+  public RemoveUnusedLocalVariableCodemod(
+      @ProvidedSonarScan(ruleId = "java:S1481") final RuleIssues issues) {
```
nahsra, line 32: We always seem to use `MATCHES_START`, so I think we should change `SonarPluginJavaParserChanger` to, by default, use `RegionNodeMatcher.MATCHES_START` without the user having to write this boilerplate every time. Then, if an individual codemod wants to change it, we can let them use this signature here to override that.

nahsra (changes requested):

core-codemods/src/main/java/io/codemodder/codemods/RemoveUnusedLocalVariableCodemod.java
carlosu7: no, i will add a test case and refactor code considering that scenario

nahsra (approved):

PR #262, opened 2023-12-14
Sonar codemod for defining a constant instead of duplicating literal string n times

carlosu7:
example https://sonarcloud.io/project/issues?open=AYu2g5It97lOFaqEjR1c&id=pixee_codemodder-java

reference https://rules.sonarsource.com/java/RSPEC-1192/

This codemod is approached in two ways based on the reported case by sonarcloud:

1. When sonar reports the message `Use already-defined constant 'MY_CONSTANT' instead of duplicating its value here` where we take the constant name inside the single quotes from the issue message to just replace it.

2. When sonar reports a message like `Define a constant instead of duplicating this literal "password" 3 times` where we need to generate a constant name, add it to the classas a field declarator and replace it at the reported locations.

**TO GENERATE A CONSTANT NAME**
We will use the string value as the constant name but there are some considerations:
* We will keep only alphanumeric characters
* We remove other characters that are not alphanumeric
* If value has no alphanumeric characters, we will take parent's node name as its name, but if there's a case where we couldn't retrieve parent's node name, `CONST` value will be taken as the constant name
* We will add a counter suffix to the variable name if there are any collisions to other declared variables in the class
* Snake case is the default naming convention we are using, but before applying it, we will look for other constants declared in class to check the naming convention used, so if camel case is used, generated constant will be camel cased

**New constant will be added as a `private` member of the class**


core-codemods/src/main/resources/io/codemodder/codemods/DefineConstantForLiteralCodemod/description.md
```diff
@@ -0,0 +1,7 @@
+TODO
+
+Our changes look something like this:
+
+```diff
+
```
carlosu7, line 6: I will add the code example soon , could you help me with the description 🙏

plugins/codemodder-plugin-sonar/src/main/java/io/codemodder/providers/sonar/api/Issue.java
```diff
@@ -33,6 +35,13 @@ public class Issue {
   @JsonProperty("message")
   private String message;
 
+  @JsonProperty("flows")
+  private List<Flow> flows;
```
carlosu7, line 39: I have added the `flows` property since inside this array we can retrieve the locations of the other duplicated strings as you can read in the sonar report file `core-codemods/src/test/resources/define-constant-for-duplicate-literal-s1192/sonar-issues.json`

core-codemods/src/main/java/io/codemodder/codemods/DefineConstantForLiteralCodemod.java
```diff
@@ -0,0 +1,153 @@
...
+    final VariableDeclarator variableDeclarator = new VariableDeclarator();
+    variableDeclarator.setInitializer(new StringLiteralExpr(stringLiteralExpr.getValue()));
+    variableDeclarator.setType(type);
+    variableDeclarator.setName(constantName);
+
+    return new FieldDeclaration(modifiers, variableDeclarator);
+  }
+
+  private String buildConstantName(final StringLiteralExpr stringLiteralExpr) {
+    return stringLiteralExpr.getValue().toUpperCase().concat("_CONSTANT");
```
carlosu7, line 151: not sure if this is the constant naming convention we want to follow...

also if would like to check if there are other variables to avoid duplicated variable names

nahsra: Gosh, no -- that's ugly. Let's create something smarter here. Feel free to design something.

Also, we need something smart here to check if there's also something with that name anywhere so we don't collide.

core-codemods/src/main/java/io/codemodder/codemods/DefineConstantForLiteralCodemod.java
```diff
@@ -0,0 +1,153 @@
...
+          Range range = node.getRange().get();
+          if (RegionNodeMatcher.MATCHES_START.matches(region, range)) {
+            replaceDuplicatedLiteralToConstantExpression(node);
+            expectedDups++;
+          }
+        }
+      }
+    }
+
+    return expectedDups == numberOfDuplications;
```
nahsra, line 86: This method is trying to do too many things at the same time. 

It's trying to count if the number of expected occurrences is what we expect. But, it's also changing them at the same time. So, what happens if we change 3, but we expected 4? We'll return `false` but we _did_ make changes -- we can only return `true` if we make changes.

Maybe we can break this into a few different methods? Collect, check the number, then invoke a second method to make the changes, if the numbers have to match up.

Why do the numbers have to match up, by the way?

core-codemods/src/main/java/io/codemodder/codemods/DefineConstantForLiteralCodemod.java
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  private void replaceDuplicatedLiteralToConstantExpression(final Node node) {
+    StringLiteralExpr stringLiteralExpr = (StringLiteralExpr) node;
+    NameExpr nameExpr = new NameExpr(buildConstantName(stringLiteralExpr));
+    stringLiteralExpr.replace(nameExpr);
+  }
+
```
nahsra, line 94: Will this be a common utility for codemods? If so, should it be in a common utility class?  I also think the name of it is wrong. It seems like it's "creating", not "finding". Let's re-think the method name.

nahsra (changes requested):
Aside from the comments left, we need more test cases to verify the behavior is what we want when we're operating on interfaces, inner classes, etc.

core-codemods/src/main/java/io/codemodder/codemods/DefineConstantForLiteralCodemod.java
carlosu7: i was thinking that changes were only applied if the `onIssueFound` method only return true, but now i see that it doesn't

core-codemods/src/main/java/io/codemodder/codemods/VariableCollector.java
```diff
@@ -0,0 +1,27 @@
...
+import com.github.javaparser.ast.visitor.VoidVisitorAdapter;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * The VariableCollector class is a visitor that traverses a {@link CompilationUnit} using the
+ * {@link VoidVisitorAdapter}. It collects variable names from VariableDeclarator nodes and stores
+ * them in a Set. The collected variables can be retrieved using the getDeclaredVariables method.
+ */
+final class VariableCollector extends VoidVisitorAdapter<Void> {
```
nahsra, line 13: Why does this need to be an adapter? Why can't we call `compilationUnit.findAll(VariableDeclarator.class)`?

Also, should we also avoid collisions with method names?  Should we be searching for any `NodeWithSimpleName`?

core-codemods/src/main/resources/io/codemodder/codemods/DefineConstantForLiteralCodemod/description.md
nahsra: You've got this

core-codemods/src/main/resources/io/codemodder/codemods/DefineConstantForLiteralCodemod/report.json
```diff
@@ -0,0 +1,7 @@
+{
+  "summary" : "Define a constant for a literal string that is duplicated n times (Sonar)",
+  "change" : "Define a constant for a literal string that is duplicated n times",
+  "references" : [
```
nahsra, line 4: Can we find another official-ish source?

core-codemods/src/test/java/io/codemodder/codemods/ConstantNameStringGeneratorTest.java
```diff
@@ -0,0 +1,174 @@
+package io.codemodder.codemods;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.HashSet;
+import java.util.Set;
+import org.junit.jupiter.api.Test;
+
```
nahsra, line 8: `final`

core-codemods/src/main/java/io/codemodder/codemods/CreateConstantForLiteral.java
```diff
@@ -0,0 +1,142 @@
...
+import com.github.javaparser.ast.expr.StringLiteralExpr;
+import com.github.javaparser.ast.nodeTypes.NodeWithSimpleName;
+import com.github.javaparser.ast.type.ClassOrInterfaceType;
+import com.github.javaparser.ast.type.Type;
+import io.codemodder.CodemodInvocationContext;
+import io.codemodder.providers.sonar.api.Issue;
+import java.util.List;
+import java.util.Optional;
+
+class CreateConstantForLiteral extends DefineConstantForLiteral {
```
nahsra, line 20: `final`

nahsra (changes requested):
I added a few tweaks here, but while I do some more thinking about the API design, I think we should think about how to verify this more strongly, given how often this will make a relatively invasive change. This combination of unit and integration tests is a good start, but before releasing it into the wild, I do feel like we need more integration tests.

Can we go run this on a few different open source projects, and use them as test cases? Can we find an OSS project that finds several (maybe dozens?) of instances of this, so we can stress test our logic here? I want to make sure not only that it works, and we don't break anything, but that our naming strategy generally produces what we want.

core-codemods/src/main/java/io/codemodder/codemods/DefineConstantForLiteral.java
```diff
@@ -0,0 +1,122 @@
...
+import io.codemodder.Position;
+import io.codemodder.RegionNodeMatcher;
+import io.codemodder.SourceCodeRegion;
+import io.codemodder.providers.sonar.api.Flow;
+import io.codemodder.providers.sonar.api.Issue;
+import io.codemodder.providers.sonar.api.TextRange;
+import java.util.*;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
```
nahsra, line 19: Need JavaDoc all over these new types. It's complicated enough to warrant.

core-codemods/src/main/java/io/codemodder/codemods/DefineConstantForLiteral.java
```diff
@@ -0,0 +1,122 @@
...
+      final CodemodInvocationContext context,
+      final CompilationUnit cu,
+      final StringLiteralExpr stringLiteralExpr,
+      final Issue issue) {
+    this.context = context;
+    this.cu = cu;
+    this.stringLiteralExpr = stringLiteralExpr;
+    this.issue = issue;
+  }
+
```
nahsra, line 39: `execute()` is such a generic name I would only expect it on a generic type, something like `Runnable` or `Thread`. What is this doing?

core-codemods/src/main/java/io/codemodder/codemods/DefineConstantForLiteralCodemod.java
```diff
@@ -0,0 +1,45 @@
...
+  @Override
+  public boolean onIssueFound(
+      final CodemodInvocationContext context,
+      final CompilationUnit cu,
+      final StringLiteralExpr stringLiteralExpr,
+      final Issue issue) {
+
+    DefineConstantForLiteral defineConstantForLiteral;
```
nahsra, line 33: This pattern suggests we probably need to compose this codemod into multiple smaller codemods, but I'm not sure that's worth the distraction at the moment.

core-codemods/src/main/java/io/codemodder/codemods/CreateConstantForLiteral.java
```diff
@@ -0,0 +1,165 @@
...
+import com.github.javaparser.ast.type.Type;
+import io.codemodder.CodemodInvocationContext;
+import io.codemodder.providers.sonar.api.Issue;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * A class that extends DefineConstantForLiteral and specializes in creating new constants for
```
nahsra, line 23: Is it always String literals?

core-codemods/src/main/java/io/codemodder/codemods/UseExistingConstantForLiteral.java
```diff
@@ -0,0 +1,44 @@
...
+
+    return message.substring(startIndex + 1, endIndex);
+  }
+
+  /**
+   * This method is not implemented in this subclass, as it assumes the constant is already defined.
+   * It remains empty in this implementation.
+   */
+  @Override
+  protected void defineConstant(final String constantName) {
```
nahsra, line 41: If this comment is true, it's not a great abstract design. We only have 2 subclasses and one of them doesn't need it.

core-codemods/src/main/resources/io/codemodder/codemods/DefineConstantForLiteralCodemod/report.json
nahsra: This still stands?

core-codemods/src/test/java/io/codemodder/codemods/DefineConstantForLiteralCodemodTest3.java
```diff
@@ -0,0 +1,12 @@
+package io.codemodder.codemods;
+
+import io.codemodder.testutils.CodemodTestMixin;
+import io.codemodder.testutils.Metadata;
+
```
nahsra, line 5: Ahh, I see we added these directories because we need different `sonar-issues.json` files? We could cram them into one `sonar-issues.json` file, but I suppose this is better test isolation.

core-codemods/src/main/java/io/codemodder/codemods/DefineConstantForLiteralCodemod.java
nahsra: This is why composition is usually preferred to inheritance!

nahsra (approved):

core-codemods/src/main/java/io/codemodder/codemods/CreateConstantForLiteral.java
carlosu7: yes

core-codemods/src/main/resources/io/codemodder/codemods/DefineConstantForLiteralCodemod/report.json
carlosu7: I didn't find any good official source other than sonarsource

PR #264, opened 2023-12-14
Sonar codemod to replace `Stream.collect(Collectors.toList())` with Java 16 + `Stream.toList()`

carlosu7:
Since Java 16 there is now a better variant to produce an unmodifiable list directly from a stream: Stream.toList().

Reference https://rules.sonarsource.com/java/RSPEC-6204/

Example: https://sonarcloud.io/project/issues?open=AYu2g5Hn97lOFaqEjR0v&id=pixee_codemodder-java

### DIFF
--- a/core-codemods/src/main/resources/io/codemodder/codemods/ReplaceStreamCollectorsToListCodemod/description.md
+++ b/core-codemods/src/main/resources/io/codemodder/codemods/ReplaceStreamCollectorsToListCodemod/description.md
@@ -1,0 +1,7 @@
+TODO
+
+Our changes look something like this:
+
+```diff
+
+```
--- a/core-codemods/src/main/resources/io/codemodder/codemods/ReplaceStreamCollectorsToListCodemod/description.md
+++ b/core-codemods/src/main/resources/io/codemodder/codemods/ReplaceStreamCollectorsToListCodemod/description.md
@@ -3,7 +3,6 @@ This change modernizes the a stream's `List` creation to be driven from the simp
 Our changes look something like this:
 
 ```diff
-     List<CodemodChange> allWeaves =
--        linesAffected.stream().map(CodemodChange::from).collect(Collectors.toList());
-+        linesAffected.stream().map(CodemodChange::from).toList();
+- List<Integer> numbers = someStream.collect(Collectors.toList());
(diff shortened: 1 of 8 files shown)


nahsra (changes requested):
This can only be enabled if we can confirm that we are on Java... 17, maybe? The `toList()` method is new. Can you add that? I believe that was something that we get from `MavenProvider`. If we can't determine the version, we should take no action.

We should make whatever changes we can to support this codemod, because I have other changes I also want to make for other rules that require know that it's at least a certain Java version.

core-codemods/src/main/resources/io/codemodder/codemods/ReplaceStreamCollectorsToListCodemod/description.md
```diff
@@ -0,0 +1,9 @@
+TODO
+
+Our changes look something like this:
+
+```diff
+     List<CodemodChange> allWeaves =
+-        linesAffected.stream().map(CodemodChange::from).collect(Collectors.toList());
```
nahsra, line 7: This is a confusing change example because it's very specific to us, can we create a more "Hello, World" example?

core-codemods/src/main/resources/io/codemodder/codemods/ReplaceStreamCollectorsToListCodemod/description.md
```diff
@@ -0,0 +1,9 @@
+TODO
```
nahsra, line 1:
```
This change modernizes the a stream's `List` creation to be driven from the simple, and more readable `Stream#toList()` method. 
```

You should make the toList a link to the Javaoc for it.

carlosu7:
> This can only be enabled if we can confirm that we are on Java... 17, maybe? The `toList()` method is new. Can you add that? I believe that was something that we get from `MavenProvider`. If we can't determine the version, we should take no action.
> 
> We should make whatever changes we can to support this codemod, because I have other changes I also want to make for other rules that require know that it's at least a certain Java version.

I have sonarlint in my IDE and checked if the rule is raised for a java project prior to 16 and it is not being triggered.

nahsra (approved):

nahsra:
Nice! Looks like we can count on Sonar to verify that for us.

PR #269, opened 2023-12-21
Sonar codemod for removing block of commented-out lines of code

carlosu7:
Reference: https://rules.sonarsource.com/java/RSPEC-125/

Docs PR https://github.com/pixee/docs/pull/120

### DIFF
--- a/core-codemods/src/main/java/io/codemodder/codemods/RemoveCommentedCodeCodemod.java
+++ b/core-codemods/src/main/java/io/codemodder/codemods/RemoveCommentedCodeCodemod.java
@@ -1,14 +1,12 @@
 package io.codemodder.codemods;
 
 import com.github.javaparser.ast.CompilationUnit;
-import com.github.javaparser.ast.Node;
 import com.github.javaparser.ast.comments.Comment;
 import io.codemodder.*;
 import io.codemodder.providers.sonar.ProvidedSonarScan;
 import io.codemodder.providers.sonar.RuleIssues;
 import io.codemodder.providers.sonar.SonarPluginJavaParserChanger;
 import io.codemodder.providers.sonar.api.Issue;
-import java.util.List;
 import javax.inject.Inject;
 
 /** A codemod for removing commented-out lines of code. */
@@ -32,7 +30,7 @@ public final class RemoveCommentedCodeCodemod extends SonarPluginJavaParserChang
   public RemoveCommentedCodeCodemod(
       @ProvidedSonarScan(ruleId = "java:S125") final RuleIssues issues) {
 
-    super(issues, Comment.class, regionNodeMatcher);
+    super(issues, Comment.class, regionNodeMatcher, NodeCollector.ALL_COMMENTS);
   }
 
   @Override
@@ -45,10 +43,5 @@ public final class RemoveCommentedCodeCodemod extends SonarPluginJavaParserChang
     comment.removeForced();
 
     return true;
-  }
-
-  @Override
-  public List<? extends Node> getAllCompilationUnitNodes(final CompilationUnit cu) {
-    return cu.getAllComments();
   }
 }
--- a/core-codemods/src/main/java/io/codemodder/codemods/RemoveCommentedCodeCodemod.java
+++ b/core-codemods/src/main/java/io/codemodder/codemods/RemoveCommentedCodeCodemod.java
@@ -1,0 +1,36 @@
+package io.codemodder.codemods;
+
+import com.github.javaparser.ast.CompilationUnit;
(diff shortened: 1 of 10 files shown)


core-codemods/src/test/resources/remove-commented-code-s125/Test.java.after
```diff
@@ -0,0 +1,128 @@
...
+    private boolean attemptWasMade = false;
+
+    public AttackResultBuilder(PluginMessages messages) {
+      this.messages = messages;
+    }
+
+    public AttackResultBuilder lessonCompleted(boolean lessonCompleted) {
+      this.lessonCompleted = lessonCompleted;
+      this.feedbackResourceBundleKey = "lesson.completed";
+      // return 0;
```
carlosu7, line 53: this comment is not being reported by sonnar, that's why it is expected to be here in the changed file

sonar report of the file: https://sonarcloud.io/project/issues?fileUuids=AYvtrjqJLCzGLicz7Abr&resolved=false&types=CODE_SMELL&id=nahsra_WebGoat_10_23

core-codemods/src/main/java/io/codemodder/codemods/RemoveCommentedCodeCodemod.java
```diff
@@ -0,0 +1,49 @@
...
+import javax.inject.Inject;
+
+/** A codemod for removing commented-out lines of code. */
+@Codemod(
+    id = "sonar:java/remove-commented-code-s125",
+    reviewGuidance = ReviewGuidance.MERGE_WITHOUT_REVIEW,
+    executionPriority = CodemodExecutionPriority.HIGH)
+public final class RemoveCommentedCodeCodemod extends SonarPluginJavaParserChanger<Comment> {
+
+  private static final RegionNodeMatcher regionNodeMatcher =
```
carlosu7, line 21: specific region node matcher written for this codemod since javaparser and sonarcloud differ on the column

core-codemods/src/main/resources/io/codemodder/codemods/RemoveCommentedCodeCodemod/description.md
```diff
@@ -0,0 +1,10 @@
+TODO
```
carlosu7, line 1: 🥺 🙏

nahsra: give it a shot, i believe in you

core-codemods/src/main/java/io/codemodder/codemods/RemoveCommentedCodeCodemod.java
nahsra: There should be a big, detailed JavaDoc on this.

core-codemods/src/main/resources/io/codemodder/codemods/RemoveCommentedCodeCodemod/report.json
```diff
@@ -0,0 +1,7 @@
+{
+  "summary" : "Removed block of commented-out lines of code (Sonar)",
+  "change" : "Removed block of commented-out lines of code",
+  "references" : [
```
nahsra, line 4: We should try to find another unbiased, expert source.

nahsra (changes requested):
Great start!

plugins/codemodder-plugin-sonar/src/main/java/io/codemodder/providers/sonar/SonarPluginJavaParserChanger.java
```diff
@@ -89,6 +88,10 @@ public boolean shouldRun() {
     return ruleIssues.hasResults();
   }
```
nahsra, line 90: I see what you're trying to do here. A few things:

1. There's no need for this to be `public`, right? Just `protected`?
2. This type hierarchy tries hard to favor composition over inheritance, but this is just straight up inheritance. Just like we invented `RegionNodeMatcher` to help codemods define how they could customize their behavior, it seems like we may need a new type so we can match that new philosophy. Maybe a new `NodeCollector` type? 

Please consider Effective Java Chapter 16 -- Favoring Composition over Inheritance, discussed here by the author: https://blogs.oracle.com/javamagazine/post/java-inheritance-composition.

3. The name is...not quite right. With that name, I'm expecting it to collect a flattened `List` of all the nodes in the AST. I think this name problem goes away when you fix the issue above anyway.

nahsra (changes requested):

nahsra (approved):

PR #281, opened 2024-02-05
Update CodeTFChange to include explicit DiffSide

drdavella:


framework/codemodder-base/src/main/java/io/codemodder/DefaultCodemodExecutor.java
```diff
@@ -274,6 +275,7 @@ private CodeTFChange translateCodemodChangetoCodeTFChange(
             codemodChange.lineNumber(),
             emptyMap(),
             changeDescription,
+            CodeTFDiffSide.LEFT,
```
drdavella, line 278: This preserves the existing behavior wrt to platform integration. But we may want to differentiate and/or switch to `RIGHT` going forward.

plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/CodeTFGenerator.java
```diff
@@ -64,8 +65,10 @@ CodeTFChangesetEntry getChanges(
 
     final String description = dependencyDescriptor.create(newDependency);
     final Map<String, String> properties = buildPropertiesMap(description);
+    // Use RIGHT side as the diff side for now, as we are only adding dependencies.
```
nahsra, line 68: I appreciate you making this change _at all_, but any chance we can add a test assertion that the Maven code is using `RIGHT`?

I could only look at this for a minute, but I think such a test would be in `MavenProviderTest.java`. Maybe we add a new line 247: that looks something like this:

```
assertThat(changes.get(0).getChanges().get(0).getSide()).isEqualTo(CodeTFDiffSide.RIGHT);
```

nahsra (changes requested):

plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/CodeTFGenerator.java
drdavella: Done. That was a pretty good breadcrumb, thanks 😅

nahsra (approved):

ryandens (approved):

PR #287, opened 2024-02-13
AutoGen documentation with Gradle

ryandens:
- :recycle: extract build logic to task


gradle/build-plugins/src/main/kotlin/io.codemodder.core-codemods-docs.gradle.kts
```diff
@@ -12,162 +10,11 @@ plugins {
...
-
-    if (codemodDocsDir.list().isEmpty()) {
-        println("Docs directory was empty")
-    }
-
-    val javaMainSources = java.sourceSets.main.get().java.srcDirs
-    val defaultCodemodSource = file("src/main/java/io/codemodder/codemods/DefaultCodemods.java").readText()
-
-    // for every codemod in the project
-    for (javaMainSource in javaMainSources) {
```
ryandens, line 39: The main important thing we want to do here is that we want to execute the task in the scope of the task action rather than at configuration time. This could technically be achieved in a `doLast {}` block but given the complexity of this task it felt right to extract it out

gradle/build-plugins/src/main/kotlin/io/codemodder/docs/GenerateDocsTask.kt
```diff
@@ -0,0 +1,192 @@
...
+    @get:Internal
+    abstract val codemodDocsDir: DirectoryProperty
+
+    @get:InputFiles
+    abstract val javaMainSources: ConfigurableFileCollection
+
+    @get:InputFile
+    abstract val defaultCodemodSource: RegularFileProperty
+
+    @Option(option = "codemodDocsDir", description = "The dir to write the docs to")
```
ryandens, line 34: the other thing this change gives us is using a command line option here, so we can run this from the command line in a bit more of a traditional way:
```bash
gw :core-codemods:foo --codemodDocsDir /Users/ryandens/git/pixee/docs/docs/codemods/java/
```

gradle/build-plugins/src/main/kotlin/io/codemodder/docs/GenerateDocsTask.kt
```diff
@@ -0,0 +1,192 @@
...
+    fun setCodemodderDocsDir(value: String) {
+        codemodDocsDir.set(project.file(value))
+    }
+
+    @get:Inject
+    abstract val projectLayout: ProjectLayout
```
ryandens, line 40: the last thing is that we use this `ProjectLayout` abstraction for accessing files during task execution, which makes it easier to be compatible with gradle configuration cache by not accessing the `project` object during task execution (e.g. in the `TaskAction` or the `doLast`

nahsra (approved):

PR #289, opened 2024-02-14
Hardened type resolution checks in QueryParameterizer

andrecsilva:
Also updated the README to mention docker for running tests.

### DIFF
--- a/core-codemods/src/main/java/io/codemodder/codemods/QueryParameterizer.java
+++ b/core-codemods/src/main/java/io/codemodder/codemods/QueryParameterizer.java
@@ -54,7 +54,7 @@ final class QueryParameterizer {
     return injections;
   }
 
-  private Optional<ResolvedType> calculateResolvedType(Expression e) {
+  private final Optional<ResolvedType> calculateResolvedType(Expression e) {
     try {
       return Optional.of(e.calculateResolvedType());
     } catch (final UnsolvedSymbolException exception) {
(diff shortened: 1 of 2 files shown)


nahsra:
Does this fix a bug? Do we have a regression test?

andrecsilva:
> Does this fix a bug? Do we have a regression test?

The SQLParameterizer codemod was crashing on this file:

https://github.com/HCL-TECH-SOFTWARE/AltoroJ/blob/1a47e993a2db5c5c4fc42b61feab03704b9ae29c/src/com/ibm/security/appscan/altoromutual/util/DBUtil.java#L212

This PR fixes that.

The core issue is with JavaParser's type resolution for some expressions. Sometimes JavaParser's type resolution will fail (and throw an exception). It happened before and I'm not entirely sure how to reproduce it, hence no tests.

It is likely this also affects other codemods that uses type resolution. I could investigate and write some tests for this issue if you feel it's worth pursuing it.

nahsra:
I've had to make similar changes in that file as you can probably see.

The code appears to be [Apache license](https://github.com/HCL-TECH-SOFTWARE/AltoroJ/blob/1a47e993a2db5c5c4fc42b61feab03704b9ae29c/LICENSE), so I suppose we could bring in a minimal version of the repo for a test case. I'll leave that up to you.

core-codemods/src/main/java/io/codemodder/codemods/QueryParameterizer.java
```diff
@@ -152,14 +171,17 @@ private boolean convertibleToString(final Expression exp) {
     // See
     // https://download.oracle.com/otn-pub/jcp/jdbc-4_2-mrel2-spec/jdbc4.2-fr-spec.pdf
     // for type conversions by setObject
-    final var type = exp.calculateResolvedType();
```
nahsra, line 155: I would do something like 

```
typeRef = calculateResolvedType(exp);
if (typeRef.isEmpty()) {
  return false;
}
var type = typeRef.get();
// now just use `type` everywhere
```

It's a cleaner diff for this PR, but it also just makes the code easier to read so we don't have `get()` scattered everywhere.

nahsra (changes requested):

core-codemods/src/main/java/io/codemodder/codemods/QueryParameterizer.java
```diff
@@ -54,7 +54,7 @@ List<Deque<Expression>> getInjections() {
     return injections;
   }
 
-  private Optional<ResolvedType> calculateResolvedType(Expression e) {
+  private final Optional<ResolvedType> calculateResolvedType(Expression e) {
```
nahsra, line 57: I meant `final` on the `Expression e` 😄

andrecsilva: Afters months of looking mostly at python code, It seems I can barely remember what `final` means.

nahsra (approved):

Issue #270, opened 2023-12-21
Processing stuck after 8 minutes


dhafley:
When running on Apache Fink, the process hangs after eight minutes. It appears to hang
[codemodder-java-jstack-output.txt](https://github.com/pixee/codemodder-java/files/13743697/codemodder-java-jstack-output.txt)
 running codemod: pixee:java/prevent-filewriter-leak-with-nio. Attached is the output from jstack

nahsra:
Can you capture 2 back to back `jstack` invocations when its hung? That may help us quickly understand if we have a deadlock.

nahsra:
Feel free to re-submit with more logs!

nahsra:
Re-opening this to comment after we saw similar behavior with Apache Pinot with similar logs to what we saw before. Keeping some notes here. 

The hangup deterministically occurs during `pixee:java/make-prng-seed-unpredictable` codemod execution from the [pixee CLI](https://github.com/pixee/pixee-cli/).

Running just this codemod on its own, it doesn't seem to get stuck:
```
./gradlew :core-codemods:run --args="--verbose --dry-run --codemod-include=pixee:java/make-prng-seed-unpredictable /tmp/code/pinot/"
```
This execution takes a minute.

The error also does not occur when I run all codemods against pinot:
```
./gradlew :core-codemods:run --args="--verbose --dry-run --codemod-include=pixee:java/make-prng-seed-unpredictable /tmp/code/pinot/"
```
This execution takes 5 minutes.

I ran both scenarios multiple times. This is an example thread dump seen during 

```
"main" #1 prio=5 os_prio=31 cpu=68362407.83ms elapsed=69739.14s tid=0x00007f946380e600 nid=0x2803 runnable  [0x000070000a46c000]
   java.lang.Thread.State: RUNNABLE
	at io.codemodder.DefaultCodemodExecutor.execute(DefaultCodemodExecutor.java:171)
	at io.codemodder.CLI.call(CLI.java:424)
	at io.codemodder.CLI.call(CLI.java:44)
...
 "pool-22-thread-1" #149 prio=5 os_prio=31 cpu=21450.33ms elapsed=69511.79s tid=0x00007f9464e64200 nid=0x942b runnable  [0x000070000b7a9000]
   java.lang.Thread.State: RUNNABLE
	at java.io.FileOutputStream.writeBytes(java.base@17.0.9/Native Method)
[...]

PR #290, opened 2024-02-16
Fixed hang

nahsra:
This (hopefully) fixes issue #270 and other reports of stalled analysis.


nahsra:
I just confirmed this also fixes scanning Flink.

framework/codemodder-base/src/main/java/io/codemodder/DefaultCodemodExecutor.java
```diff
@@ -155,17 +155,17 @@ public CodeTFResult execute(final List<Path> filePaths) {
 
             } catch (Exception e) {
```
ryandens, line 156: interesting, I think part of the problem could be that we're not preserving thread interruption state when we catch the generic `Exception` here, kinda as described in https://rules.sonarsource.com/java/tag/error-handling/RSPEC-2142/


Looking around the codebase, it looks like there are a number of places we do this which kinda makes me think we should write a codemod

ryandens (approved):
glad this moves us forward, don't want to hold this up while we figure out the longer term fixes

PR #293, opened 2024-02-29
Support scanning AppScan

nahsra:
This change introduces the ability to work on AppScan results.


gradle/build-plugins/src/main/kotlin/io.codemodder.maven-publish.gradle.kts
```diff
@@ -26,6 +26,8 @@ signing {
         val signingKey: String? by project
         val signingPassword: String? by project
         useInMemoryPgpKeys(signingKey, signingPassword)
```
nahsra, line 28: This change allows me to `./gradlew publishToMavenLocal` without worrying about signing.

andrecsilva (approved):

Issue #298, opened 2024-03-08
Switch order of literals codemod should respect previous null checks


nahsra:
There has been negative feedback to changes made by this codemod when there is a comparison to `null` directly previous to the suggested change. 

Besides looking inside the method for previous checks, we could also check to see if it's a `final` class member which has null invariance on assignment.

nahsra:
We could also respect/not report if the data came from a method parameter and has a `@NotNull` annotation on it.

PR #305, opened 2024-03-13
New conditions to not apply SwitchLiteralFirstComparisonsCodemod

carlosu7:
This codemod will not be executed if: 
1. Variable was previously initialized to a not null value 
2. Variable has a previous not null assertion 
3. Variable has a `@NotNull` / `@Nonnull `annotation

Related issue https://github.com/pixee/codemodder-java/issues/298

### DIFF
--- a/core-codemods/src/intTest/java/io/codemodder/integration/WebGoat822Test.java
+++ b/core-codemods/src/intTest/java/io/codemodder/integration/WebGoat822Test.java
@@ -95,7 +95,7 @@ final class WebGoat822Test extends GitRepositoryTest {
             .flatMap(Collection::stream)
             .collect(Collectors.toList());
 
-    assertThat(fileChanges.size(), is(56));
+    assertThat(fileChanges.size(), is(50));
 
     // we only inject into a couple files
     verifyStandardCodemodResults(fileChanges);
@@ -133,7 +133,7 @@ final class WebGoat822Test extends GitRepositoryTest {
             .flatMap(Collection::stream)
             .collect(Collectors.toList());
 
-    assertThat(fileChanges.size(), is(60));
+    assertThat(fileChanges.size(), is(54));
 
     verifyStandardCodemodResults(fileChanges);
 
--- a/core-codemods/src/main/java/io/codemodder/codemods/SwitchLiteralFirstComparisonsCodemod.java
+++ b/core-codemods/src/main/java/io/codemodder/codemods/SwitchLiteralFirstComparisonsCodemod.java
@@ -123,18 +123,21 @@ public final class SwitchLiteralFirstComparisonsCodemod
     final List<NullLiteralExpr> nullLiterals = cu.findAll(NullLiteralExpr.class);
 
     if (nullLiterals != null && !nullLiterals.isEmpty()) {
-      for (final NullLiteralExpr nullLiteralExpr : nullLiterals) {
-        if (nullLiteralExpr.getParentNode().isPresent()
-            && nullLiteralExpr.getParentNode().get() instanceof BinaryExpr parentBinaryExpr
-            && parentBinaryExpr.getOperator() == BinaryExpr.Operator.NOT_EQUALS) {
-          final Node left = parentBinaryExpr.getLeft();
-          final Node right = parentBinaryExpr.getRight();
-          if (isBinaryNodeChildEqualToSimpleName(left, name)
-              || isBinaryNodeChildEqualToSimpleName(right, name)) {
-            return true;
-          }
-        }
-      }
+      return nullLiterals.stream()
+          .filter(
+              nullLiteralExpr ->
+                  nullLiteralExpr.getParentNode().isPresent()
+                      && nullLiteralExpr.getParentNode().get() instanceof BinaryExpr)
+          .map(nullLiteralExpr -> (BinaryExpr) nullLiteralExpr.getParentNode().get())
+          .filter(
+              parentBinaryExpr -> parentBinaryExpr.getOperator() == BinaryExpr.Operator.NOT_EQUALS)
+          .anyMatch(
+              parentBinaryExpr -> {
+                final Node left = parentBinaryExpr.getLeft();
+                final Node right = parentBinaryExpr.getRight();
+                return isBinaryNodeChildPreviouslyCreatedAndEqualToSimpleName(left, name)
+                    || isBinaryNodeChildPreviouslyCreatedAndEqualToSimpleName(right, name);
+              });
     }
 
     return false;
@@ -145,7 +148,8 @@ public final class SwitchLiteralFirstComparisonsCodemod
    * {@link SimpleName}. Additionally, it verifies if the child node is located before the provided
    * {@link SimpleName}.
    */
-  private boolean isBinaryNodeChildEqualToSimpleName(final Node child, final SimpleName name) {
+  private boolean isBinaryNodeChildPreviouslyCreatedAndEqualToSimpleName(
+      final Node child, final SimpleName name) {
     return child instanceof NameExpr nameExpr
         && nameExpr.getName().equals(name)
         && isPreviousNodeBefore(name, nameExpr.getName());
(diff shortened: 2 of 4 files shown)


core-codemods/src/main/java/io/codemodder/codemods/SwitchLiteralFirstComparisonsCodemod.java
```diff
@@ -41,22 +58,228 @@ public boolean onResultFound(
...
+   */
+  private List<Node> filterNodesWithNotNullAnnotations(final List<Node> annotationNodes) {
+    final List<Node> nodesWithNotNullAnnotations = new ArrayList<>();
+    for (final Node node : annotationNodes) {
+      if (node instanceof NodeWithAnnotations<?> nodeWithAnnotations
+          && !nodeWithAnnotations.getAnnotations().isEmpty()
+          && hasNotNullOrNonnullAnnotation(nodeWithAnnotations.getAnnotations())) {
+        nodesWithNotNullAnnotations.add((Node) nodeWithAnnotations);
+      }
+    }
```
nahsra, line 221: All of these functions returning `List` should return a `List.copyOf()`, since there's no reason for them to be mutable

core-codemods/src/main/java/io/codemodder/codemods/SwitchLiteralFirstComparisonsCodemod.java
```diff
@@ -41,22 +58,228 @@ public boolean onResultFound(
...
+    return nodesWithNotNullAnnotations;
+  }
+
```
nahsra, line 224: There's a lot of repetitive list creation here in this file. Can we refactor these methods to use streams?

core-codemods/src/main/java/io/codemodder/codemods/SwitchLiteralFirstComparisonsCodemod.java
```diff
@@ -41,22 +58,228 @@ public boolean onResultFound(
...
+    // Get the arguments of the MethodCallExpr
+    final List<Node> childNodes = methodCallExpr.getChildNodes();
+
+    // Iterate through the arguments and collect NameExpr nodes
+    for (final Node node : childNodes) {
+      if (node instanceof NameExpr nameExpr) {
+        simpleNames.add(nameExpr.getName());
+      }
+    }
+
```
nahsra, line 279: Rather than use `null`, can we use `Optional`?

nahsra (changes requested):
Amazing start! Let's try to use streams more, and add a new test.

core-codemods/src/intTest/java/io/codemodder/integration/WebGoat822Test.java
```diff
@@ -133,7 +133,7 @@ void it_transforms_webgoat_with_codeql() throws Exception {
             .flatMap(Collection::stream)
             .collect(Collectors.toList());
```
nahsra, line 135: Well this is surprising -- can you see which ones we no longer find/fix?

core-codemods/src/main/java/io/codemodder/codemods/SwitchLiteralFirstComparisonsCodemod.java
```diff
@@ -238,18 +242,16 @@ private boolean isNotNullOrNonnullAnnotation(final AnnotationExpr annotation) {
...
-        final SimpleName declaratorSimpleName = declarator.getName();
-        if (declaratorSimpleName.equals(targetName)
-            && isPreviousNodeBefore(targetName, declaratorSimpleName)) {
-          final Optional<Expression> initializer = declarator.getInitializer();
-          return initializer.isPresent() && !(initializer.get() instanceof NullLiteralExpr);
-        }
-      }
-    }
-
-    return false;
```
nahsra: You could make this a little simpler/less nested/slightly less complex with:
```
.filter(declarator -> declarator.getName().equals(targetName))
.filter(declarator -> isPreviousNodeBefore(targetName, declarator.getName())

.anyMatch(declarator -> declarator.getInitializer()
                            .map(expr -> !(expr instanceof NullLiteralExpr))
                            .orElse(false));
```

I think there's some other improvements here I could make but I'm tired

core-codemods/src/intTest/java/io/codemodder/integration/WebGoat822Test.java
carlosu7: To check differences, I just ran this integration test with this codemod

Before this update, these are the files that are being modified
Screenshot 2024-03-14 at 10 17 10 p m


But now, the modified files are the following:
Screenshot 2024-03-14 at 10 17 51 p m


The 6 files that are no longer modified are:
[webgoat-lessons/cross-site-scripting/src/main/java/org/owasp/webgoat/xss/CrossSiteScriptingLesson3.java](https://github.com/WebGoat/WebGoat/blob/e75cfbeb110e3d3a2ca3c8fee2754992d89c419d/webgoat-lessons/cross-site-scripting/src/main/java/org/owasp/webgoat/xss/CrossSiteScriptingLesson3.java)

[webgoat-lessons/crypto/src/main/java/org/owasp/webgoat/crypto/SecureDefaultsAssignment.java](https://github.com/WebGoat/WebGoat/blob/e75cfbeb110e3d3a2ca3c8fee2754992d89c419d/webgoat-lessons/crypto/src/main/java/org/owasp/webgoat/crypto/SecureDefaultsAssignment.java)

[webgoat-lessons/crypto/src/main/java/org/owasp/webgoat/crypto/XOREncodingAssignment.java](https://github.com/WebGoat/WebGoat/blob/e75cfbeb110e3d3a2ca3c8fee2754992d89c419d/webgoat-lessons/crypto/src/main/java/org/owasp/webgoat/crypto/XOREncodingAssignment.java)

[webgoat-lessons/csrf/src/main/java/org/owasp/webgoat/csrf/CSRFGetFlag.java](https://github.com/WebGoat/WebGoat/blob/e75cfbeb110e3d3a2ca3c8fee2754992d89c419d/webgoat-lessons/csrf/src/main/java/org/owasp/webgoat/csrf/CSRFGetFlag.java)

[webgoat-lessons/password-reset/src/main/java/org/owasp/webgoat/password_reset/ResetLinkAssignment.java](https://github.com/WebGoat/WebGoat/blob/e75cfbeb110e3d3a2ca3c8fee2754992d89c419d/webgoat-lessons/password-reset/src/main/java/org/owasp/webgoat/password_reset/ResetLinkAssignment.java)

[webwolf/src/main/java/org/owasp/webwolf/WebWolf.java](https://github.com/WebGoat/WebGoat/blob/e75cfbeb110e3d3a2ca3c8fee2754992d89c419d/webwolf/src/main/java/org/owasp/webwolf/WebWolf.java)

carlosu7: @nahsra now that I'm checking, for the test case of :
[webwolf/src/main/java/org/owasp/webwolf/WebWolf.java](https://github.com/WebGoat/WebGoat/blob/e75cfbeb110e3d3a2ca3c8fee2754992d89c419d/webwolf/src/main/java/org/owasp/webwolf/WebWolf.java)
Former changes:
```
-        if (null==fileEncoding || !fileEncoding.equals("UTF-8")) {
+        if (null==fileEncoding || !"UTF-8".equals(fileEncoding)) {
```
`fileEncoding` is being initialized [here](https://github.com/WebGoat/WebGoat/blob/e75cfbeb110e3d3a2ca3c8fee2754992d89c419d/webwolf/src/main/java/org/owasp/webwolf/WebWolf.java#L53)

variable is initialized but it can throw a NPE (in this case it doesn't apply because of the previous `null==fileEncoding` assertion)

nahsra (approved):

nahsra:
Wait, I looked at one of the WebGoat samples, and I think we might be wrong.

See the `fistNameElement` 
example: https://github.com/WebGoat/WebGoat/blob/e75cfbeb110e3d3a2ca3c8fee2754992d89c419d/webgoat-lessons/cross-site-scripting/src/main/java/org/owasp/webgoat/xss/CrossSiteScriptingLesson3.java

Why don't we harden this case? On line 52 it gets assigned:

```
String fistNameElement = doc.select("body > table > tbody > tr:nth-child(1) > td:nth-child(2)").first().text();
```

But we have no guarantee this isn't `null`, are we being to aggressive now? I don't think we have any proof this has been initialized to a not-null value?

Issue #299, opened 2024-03-08
Hardening system command codemod hardens usage of constant data


nahsra:
If you have code that looks like either of these:

```
private static final A = "a";
Runtime.getRuntime().exec(new String[]{"a", "b", "c"});
Runtime.getRuntime().exec(new String[]{A, "b", "c"});
```

... the codemod will change it, even though it's deducible that it's hardcoded/constant.

PR #307, opened 2024-03-14
HardenProcessCreationCodemod hardens when there are any variables 

carlosu7:
Now this codemod will not be executed when the parameters received are only constant variables or hardcoded values.

Issue https://github.com/pixee/codemodder-java/issues/299

### DIFF
--- a/core-codemods/src/main/java/io/codemodder/codemods/ArgumentExpressionExtractor.java
+++ b/core-codemods/src/main/java/io/codemodder/codemods/ArgumentExpressionExtractor.java
@@ -1,0 +1,45 @@
+package io.codemodder.codemods;
+
+import com.github.javaparser.ast.Node;
+import com.github.javaparser.ast.NodeList;
+import com.github.javaparser.ast.expr.Expression;
+import com.github.javaparser.ast.expr.MethodCallExpr;
+import java.util.ArrayList;
+import java.util.List;
+
+public class ArgumentExpressionExtractor {
+
+  private ArgumentExpressionExtractor() {}
+
+  /**
+   * Extracts all expressions from the arguments and their child nodes of a MethodCallExpr.
+   *
+   * @param methodCallExpr The MethodCallExpr to extract expressions from.
+   * @return A list containing all expressions found in the MethodCallExpr arguments and their child
+   *     nodes.
+   */
+  public static List<Expression> extractExpressions(MethodCallExpr methodCallExpr) {
+    List<Expression> expressions = new ArrayList<>();
+    NodeList<Expression> arguments = methodCallExpr.getArguments();
+    for (Expression argument : arguments) {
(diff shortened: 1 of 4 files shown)


core-codemods/src/main/java/io/codemodder/codemods/HardenProcessCreationCodemod.java
```diff
@@ -52,4 +81,38 @@ public boolean onResultFound(
...
+
+  private List<FieldDeclaration> getConstantFieldDeclarations(
+      final List<FieldDeclaration> fieldDeclarations) {
+
+    final List<FieldDeclaration> constantFieldDeclarations = new ArrayList<>();
+
+    for (FieldDeclaration fieldDeclaration : fieldDeclarations) {
+      // Check if the fieldDeclaration has the 'final' modifier
+      if (fieldDeclaration.isFinal()) {
+        // Check each variable within the fieldDeclaration
```
nahsra, line 100: This is a good spot for using a `Stream` with `allMatch`:

```
boolean isAllConstant fieldDeclarations.stream().allMatch(whatever);
```

core-codemods/src/main/java/io/codemodder/codemods/ArgumentExpressionExtractor.java
```diff
@@ -0,0 +1,45 @@
...
+
+  /**
+   * Extracts all expressions from the arguments and their child nodes of a MethodCallExpr.
+   *
+   * @param methodCallExpr The MethodCallExpr to extract expressions from.
+   * @return A list containing all expressions found in the MethodCallExpr arguments and their child
+   *     nodes.
+   */
+  public static List<Expression> extractExpressions(MethodCallExpr methodCallExpr) {
+    List<Expression> expressions = new ArrayList<>();
```
nahsra, line 22: argument should be `final`

How about:
```
methodCallExpr.getArguments().stream().flatten(..).map(..to expression..).toList();
```

nahsra (changes requested):
Some of the same feedback from the last review applies here as well. Let's try to use streams more!

core-codemods/src/main/java/io/codemodder/codemods/ArgumentExpressionExtractor.java
```diff
@@ -0,0 +1,45 @@
...
+
+  private ArgumentExpressionExtractor() {}
+
+  /**
+   * Extracts all expressions from the arguments and their child nodes of a MethodCallExpr.
+   *
+   * @param methodCallExpr The MethodCallExpr to extract expressions from.
+   * @return A list containing all expressions found in the MethodCallExpr arguments and their child
+   *     nodes.
+   */
```
nahsra, line 20: If the class is package protected, no need for this to be `public`

nahsra (approved):

PR #308, opened 2024-03-15
SwitchLiteralFirstComparisonsCodemod - check that variable initializer is a string

carlosu7:
In order to avoid assuming that given initializer for the string variable is a not null value, we check that initializer is a StringLiteralExpr


nahsra:
Don't hate me, but I slept on this and I think I want to get user feedback on the way the first design you proposed before we take this step.

PR #312, opened 2024-03-20
Add detection tool metadata to CodeTF

drdavella:
This adds the `detectionTool` metadata field to the CodeTF result and populates the appropriate `detectionTool` metadata for Sonar codemods. I will create separate issues for updating CodeQL and Semgrep codemods.

### DIFF
--- a/framework/codemodder-base/src/main/java/io/codemodder/CodeChanger.java
+++ b/framework/codemodder-base/src/main/java/io/codemodder/CodeChanger.java
@@ -2,7 +2,6 @@ package io.codemodder;
 
 import io.codemodder.codetf.CodeTFReference;
 import io.codemodder.codetf.DetectionTool;
-
 import java.nio.file.Path;
 import java.util.List;
 
--- a/framework/codemodder-base/src/main/java/io/codemodder/CodeChanger.java
+++ b/framework/codemodder-base/src/main/java/io/codemodder/CodeChanger.java
@@ -1,6 +1,8 @@
 package io.codemodder;
 
 import io.codemodder.codetf.CodeTFReference;
+import io.codemodder.codetf.DetectionTool;
+
 import java.nio.file.Path;
 import java.util.List;
 
@@ -12,6 +14,11 @@ public interface CodeChanger {
 
   /** A deep description of what this codemod's changes. */
   String getDescription();
+
+  /** Detection tool metadata for this codemod. */
+  default DetectionTool getDetectionTool() {
+    return null;
+  }
 
   /**
    * A list of references for further reading on the issues this codemod addresses or other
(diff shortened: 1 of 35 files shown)


drdavella:
See #314 and #313.

nahsra (approved):

framework/codemodder-base/src/main/java/io/codemodder/CodeChanger.java
```diff
@@ -13,6 +14,11 @@ public interface CodeChanger {
   /** A deep description of what this codemod's changes. */
   String getDescription();
 
+  /** Detection tool metadata for this codemod. */
+  default DetectionTool getDetectionTool() {
+    return null;
```
ryandens, line 19: rather than have this be null, is there a reason to not have all codemods implement this and specify if/when pixee analysis was the detection tool?

nahsra: That's true. This sort of implies pixee is special and the implied value when its `null`, and that doesn't feel right.

drdavella: We could do this. But the original intent of this field was to support metadata about rules from _external_ detection tools. For one thing, the spec indicates that this field is strictly optional. And my expectation is that this field is going to be used by upstream tools to render the results of SAST remediation. So even if we populate this for Pixee codemods it is going to be ignored for those cases.

There's no reason we couldn't change our minds about this going forward though.

Fwiw I wanted to use `Optional` instead of null here but it didn't play nicely with our serializers.

ryandens: Check out #315 -  I'm good with either defining Pixee detection tools as well, but ok with either approach. From an implementation perspective, I prefer leaving `getDetectionTool` off of `CodeChanger`. Since its optional for codemods to implement, I would rather have `CodeChanger`s optionally mixin the interface for being a fix-only codemod than to optionally override a default implementation that returns null. 

At the end of the day, this is far from the biggest deal and don't want to hold up this functionality making it in

ryandens (approved):

plugins/codemodder-plugin-sonar/src/main/java/io/codemodder/providers/sonar/SonarPluginJavaParserChanger.java
```diff
@@ -93,6 +96,13 @@ public boolean shouldRun() {
     return ruleIssues.hasResults();
   }
 
+  protected abstract DetectorRule getDetectorRule();
+
+  @Override
+  public DetectionTool getDetectionTool() {
+    return new DetectionTool("Sonar", getDetectorRule(), List.of());
```
ryandens, line 103: I just noticed the empty list of `findings` here - any plans to have this return a non empty list?

drdavella: Yes we want to do that eventually. I think we'll allow that to be driven by the needs of CodeTF consumers.

PR #315, opened 2024-03-21
:ok_hand: detection tool feedback

ryandens:
- **:ok_hand: Define FixOnlyCodeChanger interface and make sonar codemods implement it**

This is really just a different way of accomplishing the very same thing as you did, but maps more to how I envision evolving the internal API for writing codemods in Java


drdavella (approved):
Makes sense to me. I didn't realize this kind of introspection is available and/or idiomatic in Java but I'll defer to your judgment 😅

ryandens:
There's pros and cons to the instanceof check, but in this case it seems reasonable

PR #317, opened 2024-03-21
Xstream use deny types instead of register converter when possible

carlosu7:


core-codemods/src/main/java/io/codemodder/codemods/HardenXStreamCodemod.java
```diff
@@ -52,9 +94,52 @@ private static Statement buildFixStatement(final String variableName) {
...
+  private Optional<DependencyGAV> getXstreamDependency(final CodemodInvocationContext context) {
+    return context.dependencies().stream()
+        .filter(
+            dependency ->
+                "com.thoughtworks.xstream".equals(dependency.group())
+                    && "xstream".equals(dependency.artifact()))
+        .findFirst();
+  }
+
+  private static boolean isGreaterThanOrEqualTo(String version, String compareToVersion) {
```
nahsra, line 106: This seems like it should be in a common utility somewhere and be test cased to hell

framework/codemodder-base/src/test/java/io/codemodder/DefaultCodemodExecutorTest.java
```diff
@@ -142,6 +142,11 @@ public DependencyUpdateResult updateDependencies(
       return DependencyUpdateResult.create(
           remainingFileDependencies, skippedDependencies, changes, Set.of());
     }
+
+    @Override
+    public Collection<DependencyGAV> getAllDependencies(Path projectDir, Path file) {
+      return null;
```
nahsra, line 148: Why not `List.of()`?

core-codemods/src/main/java/io/codemodder/codemods/HardenXStreamCodemod.java
```diff
@@ -52,9 +94,52 @@ private static Statement buildFixStatement(final String variableName) {
...
+            Integer.parseInt(versionParts[2]), Integer.parseInt(compareToVersionParts[2]));
+    return patchVersionComparison >= 0;
+  }
+
   @Override
   public List<DependencyGAV> dependenciesRequired() {
-    return List.of(JAVA_SECURITY_TOOLKIT_XSTREAM);
+
+    return useDenyTypes
+        ? List.of(DependencyGAV.JAVA_SECURITY_TOOLKIT)
```
nahsra, line 141: Yeah, this needs a refactor. The dependencies required are not static. We have two options:

1. Refactor this into 2 codemods, one of which uses `denyTypes` one which uses `registerConverter`. We can compose them like `HardenJavaDeserializationCodemod`.
2. Refactor `dependenciesRequired()` and its callers so that it is dynamic, according to the change being made.

I think 2 feels like the correct solution.

nahsra (commented):
Left feedback

nahsra:
Closing this until it's ready for another review

PR #323, opened 2024-03-28
Refactor  `JavaParserChanger` onResultFound and onIssueFound to return an object

carlosu7:
This refactor was done in order to provide a `ChangesResult` object that will contain a boolean that indicates if changes were done by the codemod and the list of dependencies added by it


framework/codemodder-base/src/main/java/io/codemodder/javaparser/ChangesResult.java
```diff
@@ -0,0 +1,64 @@
+package io.codemodder.javaparser;
+
+import io.codemodder.DependencyGAV;
+import java.util.List;
+
+/** Represents the result of changes made during parsing. */
+public class ChangesResult {
```
nahsra, line 7: I believe this this should be an `interface` (named `ChangeResults`).

It should offer instances like:

```java
public interface ChangesResult {

   // the methods

  // easy instances to re-use
  ChangeResults noChanges = new DefaultChangeResults(false, List.of());

  // a default builder
 static ChangeResults from(boolean changesApplied, List dependenciesRequired) {
  ...
 }
```

Also we should null check the dependencies because I hate having to deal with null values.

framework/codemodder-base/src/main/java/io/codemodder/javaparser/ChangesResult.java
```diff
@@ -0,0 +1,64 @@
...
+  public static ChangesResult changesApplied(final List<DependencyGAV> dependenciesRequired) {
+    return new ChangesResult(true, dependenciesRequired);
+  }
+
+  /**
+   * Checks if changes were applied.
+   *
+   * @return True if changes were applied, false otherwise.
+   */
+  public boolean areChangesApplied() {
```
nahsra, line 52: This prompts some thinking. The _reason_ changes are not applied may be interesting to the caller, but maybe we don't need to expose this yet. I'd love to know "not changed because shape was unexpected" or "not changed because I failed".

framework/codemodder-base/src/main/java/io/codemodder/javaparser/ChangesResult.java
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  /**
+   * Creates a ChangesResult indicating that changes were applied and additional dependencies are
+   * required.
+   *
+   * @param dependenciesRequired List of dependencies required due to changes applied.
+   * @return ChangesResult instance with changes applied and additional dependencies required.
+   */
+  public static ChangesResult changesApplied(final List<DependencyGAV> dependenciesRequired) {
```
nahsra, line 43: What do you think about renaming this to `changesAppliedWith()`?

nahsra (changes requested):
Some small changes to consider!

framework/codemodder-base/src/main/java/io/codemodder/javaparser/ChangesResult.java
carlosu7: a new string field `reason` to be used by the caller when no changes were applied?

should we create another ticket for it?

framework/codemodder-base/src/main/java/io/codemodder/javaparser/DefaultChangesResult.java
```diff
@@ -0,0 +1,25 @@
+package io.codemodder.javaparser;
+
+import io.codemodder.DependencyGAV;
+import java.util.List;
+
```
nahsra, line 5: `final`

nahsra (approved):

Issue #303, opened 2024-03-08
Investigate SQL parameterizer cases


nahsra:
These locations are reported by a scanner as being vulnerable to SQLi for WebGoat 8.2.3. We should investigate if any of these offer opportunities for improvement of our fix logic.

org.owasp.webgoat.container.users.UserService:52
org.owasp.webgoat.lessons.challenges.challenge5.Assignment5:59
org.owasp.webgoat.lessons.sqlinjection.advanced.SqlInjectionChallenge:69
org.owasp.webgoat.lessons.sqlinjection.advanced.SqlInjectionLesson6a:74
org.owasp.webgoat.lessons.sqlinjection.introduction.SqlInjectionLesson10:71
org.owasp.webgoat.lessons.sqlinjection.introduction.SqlInjectionLesson2:65
org.owasp.webgoat.lessons.sqlinjection.introduction.SqlInjectionLesson3:63
org.owasp.webgoat.lessons.sqlinjection.introduction.SqlInjectionLesson4:62
org.owasp.webgoat.lessons.sqlinjection.introduction.SqlInjectionLesson5a:67
org.owasp.webgoat.lessons.sqlinjection.introduction.SqlInjectionLesson5b:65
org.owasp.webgoat.lessons.sqlinjection.introduction.SqlInjectionLesson8:78
org.owasp.webgoat.lessons.sqlinjection.introduction.SqlInjectionLesson8:158
org.owasp.webgoat.lessons.sqlinjection.introduction.SqlInjectionLesson9:76
org.owasp.webgoat.lessons.sqlinjection.introduction.SqlInjectionLesson9:76
org.owasp.webgoat.lessons.sqlinjection.mitigation.Servers:72

andrecsilva:
These were correctly being parameterized.

```
org.owasp.webgoat.lessons.sqlinjection.introduction.SqlInjectionLesson8:158
org.owasp.webgoat.lessons.sqlinjection.advanced.SqlInjectionChallenge:69
```

After PRs #319 and #324, the examples below are now being parameterized.

```
org.owasp.webgoat.lessons.sqlinjection.introduction.SqlInjectionLesson10:71
org.owasp.webgoat.lessons.sqlinjection.advanced.SqlInjectionLesson6a:74
org.owasp.webgoat.lessons.sqlinjection.introduction.SqlInjectionLesson5a:67
org.owasp.webgoat.lessons.sqlinjection.introduction.SqlInjectionLesson8:78
[...]

PR #324, opened 2024-03-29
Improvements to the parameterizer codemods

andrecsilva:
Changed the testing conditions for the SQL/HQL parameterizer codemod to make changes.

- Weakened conditions for making changes. The algorithm will make changes to variables that are used after declaration. This may cause some wrong changes to be made, but those are rare.
- Refactored `LinearizeStringExpression` out of `QueryParameterizer`. This may help to write other injection based codemods.
- Added transformations for removing empty string concatenations and unused variables. Those are meant to be used as a cleanup step to the parameterizer codemods, but the last one is not currently in use due to a (just discovered) JavaParser bug. This means the transformation may leave some unused variables in the code.

These changes make it possible to fix the following examples listed in #303.
```
org.owasp.webgoat.lessons.sqlinjection.advanced.SqlInjectionLesson6a:74
org.owasp.webgoat.lessons.sqlinjection.introduction.SqlInjectionLesson5a:67
org.owasp.webgoat.lessons.sqlinjection.introduction.SqlInjectionLesson8:78
org.owasp.webgoat.lessons.sqlinjection.introduction.SqlInjectionLesson9:76
```


framework/codemodder-base/src/main/java/io/codemodder/ast/ASTs.java
```diff
@@ -660,6 +660,9 @@ public static boolean isNotInitializedAndAssignedAtMostOnce(LocalVariableDeclara
    * Returns a {@link List} containing all the referenced of {@code localDeclaration} in its scope.
    */
   public static List<NameExpr> findAllReferences(final LocalVariableDeclaration localDeclaration) {
```
nahsra, line 662: Remove printlns?

framework/codemodder-base/src/main/java/io/codemodder/ast/LinearizedStringExpression.java
```diff
@@ -0,0 +1,133 @@
...
+
+  public Map<Expression, Expression> getResolvedExpressionsMap() {
+    return this.resolvedExpressions;
+  }
+
+  /**
+   * Linearizes a string expression. By linearization, it means that if a string is composed by the
+   * concatenation of several string literals and expressions, it returns all the expressions in
+   * order of concatenation.
+   */
```
nahsra, line 44: All other args are `final` in this type so we should probably match

framework/codemodder-base/src/test/java/io/codemodder/ast/ASTsTest.java
```diff
@@ -674,6 +674,51 @@ void it_calculates_scope_of_lambda_parameter() {
...
+				}
+			}
+		""";
+    var cu = new JavaParser().parse(code).getResult().get();
+    var lvd =
+        LocalVariableDeclaration.fromVariableDeclarator(cu.findAll(VariableDeclarator.class).get(0))
+            .get();
+    assertThat(ASTs.findAllAssignments(lvd).toList().size(), is(1));
+  }
+
```
nahsra, line 701: Awesome!

core-codemods/src/main/java/io/codemodder/codemods/HQLParameterizationCodemod.java
```diff
@@ -43,9 +42,15 @@ private Optional<CodemodChange> onNodeFound(
   @Override
   public List<CodemodChange> visit(
       final CodemodInvocationContext context, final CompilationUnit cu) {
-    return cu.findAll(MethodCallExpr.class).stream()
-        .flatMap(mce -> onNodeFound(context, mce, cu).stream())
-        .collect(Collectors.toList());
+    final var allChanges =
+        cu.findAll(MethodCallExpr.class).stream()
+            .flatMap(mce -> onNodeFound(context, mce, cu).stream())
+            .collect(Collectors.toList());
```
nahsra, line 48: This seems like too broad of a scope to apply this `removeEmptyString`. Can we limit to the `BlockStmt` or method in which we found this? I'm worried about making unrelated changes and confusing the user.

nahsra (changes requested):

core-codemods/src/main/java/io/codemodder/codemods/SQLParameterizerCodemod.java
```diff
@@ -17,6 +19,12 @@ public final class SQLParameterizerCodemod extends JavaParserChanger {
 
   private Optional<CodemodChange> onNodeFound(final MethodCallExpr methodCallExpr) {
     if (new SQLParameterizer(methodCallExpr).checkAndFix()) {
+      var maybeMethodDecl = methodCallExpr.findAncestor(MethodDeclaration.class);
+      // Cleanup, removes empty string concatenations and unused variables
+      maybeMethodDecl.ifPresent(md -> ASTTransforms.removeEmptyStringConcatenation(md));
+      // TODO hits a bug with javaparser, where adding nodes won't result in the correct children
```
nahsra, line 25: What is the TODO here? Should we create a ticket for ourselves to fix this once the upstream bug is fixed?

core-codemods/src/main/java/io/codemodder/codemods/SQLParameterizerCodemod.java
```diff
@@ -17,6 +19,12 @@ public final class SQLParameterizerCodemod extends JavaParserChanger {
 
   private Optional<CodemodChange> onNodeFound(final MethodCallExpr methodCallExpr) {
     if (new SQLParameterizer(methodCallExpr).checkAndFix()) {
+      var maybeMethodDecl = methodCallExpr.findAncestor(MethodDeclaration.class);
```
nahsra, line 22: This could be a `ConstructorDeclaration` too, right?

core-codemods/src/main/java/io/codemodder/codemods/SQLParameterizerCodemod.java
andrecsilva: The TODO is mostly there to remember to uncomment this once the issue is solved or we find a workaround. I'll create both a ticket here and upstream for this bug. For now, I'm investigating a bit more to gather better feedback or to solve it myself.

core-codemods/src/main/java/io/codemodder/codemods/SQLParameterizerCodemod.java
andrecsilva: Good call. I'll add it.

framework/codemodder-base/src/test/java/io/codemodder/ast/TestUtils.java
```diff
@@ -0,0 +1,18 @@
...
+
+import com.github.javaparser.JavaParser;
+import com.github.javaparser.ParserConfiguration;
+import com.github.javaparser.ast.CompilationUnit;
+import com.github.javaparser.symbolsolver.JavaSymbolSolver;
+import com.github.javaparser.symbolsolver.resolution.typesolvers.ReflectionTypeSolver;
+
+final class TestUtils {
+
+  static CompilationUnit parseCode(final String code) {
```
nahsra, line 11: We should use `JavaParserFactory` instead.

framework/codemodder-base/src/test/java/io/codemodder/ast/TestUtils.java
```diff
@@ -0,0 +1,18 @@
+package io.codemodder.ast;
+
+import com.github.javaparser.JavaParser;
+import com.github.javaparser.ParserConfiguration;
+import com.github.javaparser.ast.CompilationUnit;
+import com.github.javaparser.symbolsolver.JavaSymbolSolver;
+import com.github.javaparser.symbolsolver.resolution.typesolvers.ReflectionTypeSolver;
+
```
nahsra, line 8: Can we scope down the type name here? The scope of it is so broad.

nahsra (changes requested):

PR #326, opened 2024-04-02
Adds remove empty string concatenation transform after sql parameterizer transform call in test

andrecsilva:
The removal of empty strings after parameterization by `SQLParameterizer` is now done as a separate transform in `ASTs.RemoveEmptyStringConcatenation`. This PR adds this transformation after the parameterization in the newly added test.


core-codemods/src/test/resources/defectdojo-sql-injection/SqlInjectionLesson8/SqlInjectionLesson8.java.after
```diff
@@ -154,7 +151,7 @@ query,                 ResultSet.TYPE_SCROLL_INSENSITIVE, ResultSet.CONCUR_UPDAT
     String time = sdf.format(cal.getTime());
 
     String logQuery =
-        "INSERT INTO access_log (time, action) VALUES (?" + "" + ", ?" + "" + ")";
```
nahsra, line 157: Looks better! Is there any chance of combining these into one string, e.g., `"(?, ?)"`

nahsra (approved):

Issue #327, opened 2024-04-02
Bundle together SQLParameterization, RemoveEmptyStringsConcatenation and RemoveUnusedVariables into a single transform.


andrecsilva:
The SQLParameterizer may require cleanup tasks since it can result in empty strings and unused variables. Those are available as separate transforms in ASTTransforms.

As of PR  #325, SQLParameterization transform is now used in several codemods. This can lead to duplicate code as we can see in #326.

PR #329, opened 2024-04-03
Extra cleanup for SQLParameterizer

### LINKED ISSUE #300: SQL parameterization codemod puts together weird strings
This codemod can makes a change to WebGoat that is difficult to read:

```diff
- "INSERT INTO access_log (time, action) VALUES ('" + time + "', '" + action "')";
+ INSERT INTO access_log (time, action) VALUES (?" + ", ?" + ")";
```

This change should be:

```bash
INSERT INTO access_log (time, action) VALUES (?, ?)
```

Current WebGoat `main` at time of writing has this issue. I think the change is functionally correct, but it’s so awkward to read that it was tripping a user up.

andrecsilva:
Adds:
- A new transformation for replacing concatenations of string literals with a single literal.
- The aforementioned transformation as a cleanup step for SQLParameterizer.
- A new transformation that combines the SQLParamterizer with the extra cleanup steps.

Closes #300 and #327.

### DIFF
--- a/core-codemods/src/main/java/io/codemodder/codemods/SQLParameterizerWithCleanup.java
+++ b/core-codemods/src/main/java/io/codemodder/codemods/SQLParameterizerWithCleanup.java
@@ -1,0 +1,28 @@
+package io.codemodder.codemods;
+
+import com.github.javaparser.ast.body.CallableDeclaration;
+import com.github.javaparser.ast.expr.MethodCallExpr;
+import io.codemodder.ast.ASTTransforms;
+
+public final class SQLParameterizerWithCleanup {
+
+  public static boolean checkAndFix(final MethodCallExpr methodCallExpr) {
+    var maybeFixed = new SQLParameterizer(methodCallExpr).checkAndFix();
+    if (maybeFixed.isPresent()) {
+      // Cleanup
+      var maybeMethodDecl = methodCallExpr.findAncestor(CallableDeclaration.class);
+      // Remove concatenation with empty strings e.g "first" +  "" -> "first";
+      maybeMethodDecl.ifPresent(cd -> ASTTransforms.removeEmptyStringConcatenation(cd));
+      // TODO hits a bug with javaparser, where adding nodes won't result in the correct children
+      // order. This causes the following to remove actually used variables
+      // maybeMethodDecl.ifPresent(md -> ASTTransforms.removeUnusedLocalVariables(md));
+
+      // Merge concatenated literals, e.g. "first" + " and second" -> "first and second"
+      maybeFixed
+          .flatMap(mce -> mce.getArguments().getFirst())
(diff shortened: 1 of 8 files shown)


core-codemods/src/main/java/io/codemodder/codemods/SQLParameterizerWithCleanup.java
```diff
@@ -0,0 +1,30 @@
+package io.codemodder.codemods;
+
+import com.github.javaparser.ast.body.CallableDeclaration;
+import com.github.javaparser.ast.expr.MethodCallExpr;
+import io.codemodder.ast.ASTTransforms;
+
```
nahsra, line 6: If this is the `public` thing, should the other thing become `package-private`? Does the user need to know/care that this is `WithCleanup`? As a caller of this API, it makes me wonder if I am calling the right thing. I want a very simple entry point.

core-codemods/src/test/resources/defectdojo-sql-injection/SqlInjectionLesson8/SqlInjectionLesson8.java.after
```diff
@@ -151,7 +150,7 @@ query,                 ResultSet.TYPE_SCROLL_INSENSITIVE, ResultSet.CONCUR_UPDAT
     String time = sdf.format(cal.getTime());
 
     String logQuery =
```
nahsra, line 152: Awesome!

framework/codemodder-base/src/main/java/io/codemodder/ast/ASTTransforms.java
```diff
@@ -291,4 +292,75 @@ public static void removeUnusedLocalVariables(final Node subtree) {
...
+      var right = binExpr.asBinaryExpr().getRight().asStringLiteralExpr();
+      binExpr.replace(binExpr.getLeft());
+      return Optional.of(right);
+    }
+    if (binExpr.isStringLiteralExpr()) {
+      return Optional.of(binExpr.asStringLiteralExpr());
+    }
+    return Optional.empty();
+  }
+
```
nahsra, line 308: This seems like this will be very useful!

nahsra (changes requested):
This is fantastic. My only issue is I think we may exposing too many public APIs unncessarily.

core-codemods/src/main/java/io/codemodder/codemods/SQLParameterizerWithCleanup.java
andrecsilva: Good point. I think there is some utility in exposing it, but I can see why having both can cause some confusion.

I can see the other being used in custom codemods. This could be useful if you want to do some additional transformation over the object (e.g. try to wrap it in a try statement) or if you simply would like your own linters to deal with the cleanup like removing the unused variables.

Issue #316, opened 2024-03-21
Improve switch-literal-order codemod to recognize non-null sources


nahsra:
Here's an example [1] of the codemod acting on the result of `String#split()`, which, AFAICT, can't return any `null` values. Therefore, there's no risk of the code throwing NPE. It would be great to create a `List` of these in the codemod so they're easy to add:

```
List<String> commonMethodsThatCantReturnNull = List.of("java.lang.String#split");
```

[1] https://github.com/nahsra/Calendar-PhoneGap-Plugin/pull/1/files#diff-2752dec9f2403b0ac18cbb21c9d8068d8623a2b024e27ba50bf8a3225c6ef8c3R380

PR #330, opened 2024-04-04
Improve switch-literal-order codemod to recognize non-null sources

carlosu7:
Method `isSimpleNameANotNullInitializedVariableDeclarator`  was updated to review expression when is a method call expression.

This new update checks if method used in the expression belongs to a library that is considered a null safe imported library (based on a given list of null safe libraries).

Issue https://github.com/pixee/codemodder-java/issues/316


core-codemods/src/main/java/io/codemodder/codemods/SwitchLiteralFirstComparisonsCodemod.java
```diff
@@ -269,6 +270,141 @@ private Optional<SimpleName> getSimpleNameFromMethodCallExpr(
...
+    final List<VariableDeclarator> variableDeclarators = cu.findAll(VariableDeclarator.class);
+    return variableDeclarators.stream()
+        .filter(declarator -> declarator.getName().equals(simpleName))
+        .filter(declarator -> isPreviousNodeBefore(simpleName, declarator.getName()))
+        .findFirst();
+  }
+
   private static final Set<String> flippableComparisonMethods =
       Set.of("equals", "equalsIgnoreCase");
+
```
nahsra, line 403: Should this test thing be there?

core-codemods/src/main/java/io/codemodder/codemods/SwitchLiteralFirstComparisonsCodemod.java
```diff
@@ -269,6 +270,141 @@ private Optional<SimpleName> getSimpleNameFromMethodCallExpr(
...
+  private static final String TEST_IMPLICIT_IMPORT = "io.codemodder.codemods.Codemod#printMessage";
+  private static final List<String> commonMethodsThatCantReturnNull =
+      List.of(
+          TEST_IMPLICIT_IMPORT,
+          "java.lang.String#replace",
```
nahsra, line 408: I would like to all the methods from `String` that can't return `null`. Like:
* `replaceAll`/`replaceFirst`
* `split`
* `toLowerCase` / `toUpperCase`
* `strip`
* `join` (might need to verify this one)
* `substring`

nahsra (changes requested):

PR #332, opened 2024-04-05
:wrench: configure nexus plugin to try 100 times rather than 60 times due to timeouts waiting for resository to transition

ryandens:
See https://github.com/gradle-nexus/publish-plugin?tab=readme-ov-file#retries-for-state-transitions

There isn't a "timeout" exactly but rather a backoff and a max number of times that can be configured for awaiting repository state change. I didn't see a reason to change the `delayBetween` value because it won't impact the CI minutes used when releasing, but if we try more times we effectively increase the length of time we'll wait for the repository to transition

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,3 +1,5 @@
+import java.time.Duration
+
 plugins {
     alias(buildlibs.plugins.nexus.publish)
     id("io.codemodder.base")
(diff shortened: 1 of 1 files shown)


nahsra (approved):
Approved, but is there also a way to space out the retries? I thought I saw some chatter about that on some threads.

nahsra (approved):

Issue #334, opened 2024-04-05
Update verboseRequestMapping codemod to delete unused RequestMapping imports


erosiec:
[pixee:java/verbose-request-mapping](https://docs.pixee.ai/codemods/java/pixee_java_verbose-request-mapping/) replaces all uses of `RequestMapping` with shortcut variants, so unused `RequestMapping` imports can be deleted. 

**example:** [https://github.com/davewichers/TestGitHubAdvSec/pull/7](https://github.com/davewichers/TestGitHubAdvSec/pull/7) 

In this case, two required imports were added: 

`+ import org.springframework.web.bind.annotation.GetMapping;`
`+ import org.springframework.web.bind.annotation.PostMapping;`

so this import can be deleted: 

`import org.springframework.web.bind.annotation.RequestMapping;`

Reported by: @davewichers

davewichers:
`import org.springframework.web.bind.annotation.RequestMethod;` can be removed too, as all uses of it were replaced as well.

PR #336, opened 2024-04-08
Add condition to removeUnusedImports and add removeUnusedImports to VerboseRequestMappingCodemod

carlosu7:
Issue https://github.com/pixee/codemodder-java/issues/334


framework/codemodder-base/src/main/java/io/codemodder/ast/ASTTransforms.java
```diff
@@ -215,6 +216,12 @@ public static void removeImportIfUnused(final CompilationUnit cu, final String c
         .anyMatch(n -> n.getNameAsString().equals(simpleName))) {
       return;
     }
+    if (cu.findAll(Node.class).stream()
```
carlosu7, line 219: Before this change annotation expression nodes were not considered since they are not NodeWithSimpleName, instead they're NodeWithName

nahsra (approved):

Issue #338, opened 2024-04-09
LexicalPreservingPrinter will remove comments when using `ASTTransforms.wrapIntoResource`


andrecsilva:
Consider the following snippet:
```java
    String code =
        """
    import java.nio.file.Files;
    import java.io.File;
    import java.io.BufferedWriter;
    class A{
	    void f(File f){
		    // first comment
		    BufferedWriter writer = Files.newBufferedWriter(f.toPath());
		    // second comment
		    System.out.println(1);
		    // third comment
		    System.out.println(2);
	    }
    }
    """;
    ParserConfiguration parserConfiguration = new ParserConfiguration();
    parserConfiguration.setLanguageLevel(ParserConfiguration.LanguageLevel.BLEEDING_EDGE);
    parserConfiguration.setSymbolResolver(new JavaSymbolSolver(new ReflectionTypeSolver()));
    var parser = new JavaParser(parserConfiguration);

    CompilationUnit cu = parser.parse(code).getResult().get();
    LexicalPreservingPrinter.setup(cu);

    var node = cu.findAll(VariableDeclarator.class).get(0);
    var lvd = LocalVariableDeclaration.fromVariableDeclarator(node).get();
    ASTTransforms.wrapIntoResource(lvd.getStatement().asExpressionStmt(), lvd.getVariableDeclarationExpr(), lvd.getScope());

    System.out.println(cu);
    System.out.println("------------------");
    System.out.println(LexicalPreservingPrinter.print(cu));
```

This will output:
```
    import java.nio.file.Files;
    import java.io.File;
    import java.io.BufferedWriter;

    class A {

        void f(File f) {
            // first comment
            try (BufferedWriter writer = Files.newBufferedWriter(f.toPath())) {
                // second comment
                System.out.println(1);
                // third comment
                System.out.println(2);
            }
        }
    }

    ------------------
    import java.nio.file.Files;
    import java.io.File;
    import java.io.BufferedWriter;
    class A{
     void f(File f){
  
      try (BufferedWriter writer = Files.newBufferedWriter(f.toPath())) {
          System.out.println(1);
          System.out.println(2);
      }


     }
    }
```

Notice how the comments disappears if we print it with the `LexicalPreservingPrinter`, despite being there if we chose not to use it.

This affect all the codemods that uses this transform, mainly those using the `ResourceLeakFixer` transform:
[...]

PR #339, opened 2024-04-09
Added resource leak transform and general codemod

### LINKED ISSUE #304: BufferedWriter migration codemod text has incorrect claims
There are benefits to the I/O performance and to the modernization of the change, but the codemod summary and documentation goes overboard in claims about how it helps. IIRC, there was a phased approach to this codemod -- first, modernize, second, wrap with try-catch/try-with-resources, and the documentation overstates the phase of where the codemod is. These need to match exactly.

The scope of this ticket is to reset the claims to match the codemod, or to improve the codemod to wrap with try-related protections. We should be able to compose this feature into the codemod [from this codemod](https://docs.pixee.ai/codemods/java/codeql_java_input-resource-leak).

Credit: Joe Beeton, Contrast

andrecsilva:
Added a new resource leak transform and general codemod

- Removed CodeQL dependency for ResourceLeakFixer and left it available as a general transform.
- Added new resource leak codemod independent of CodeQL. This may conflict with the resource leak codemods if both are available, producing more changes than necessary.
- Added ResourceLeakFixer transform to PreventFileWriterLeakWithFilesCodemod. Closes #304.

You may notice the changes to the PreventFileWriterLeakWithFilesCodemod test files. This is due to a newly discovered issue. Refer to #338.

### DIFF
--- a/core-codemods/src/intTest/java/io/codemodder/integration/WebGoat822Test.java
+++ b/core-codemods/src/intTest/java/io/codemodder/integration/WebGoat822Test.java
@@ -117,7 +117,6 @@ final class WebGoat822Test extends GitRepositoryTest {
           outputFile.getPath(),
           "--sarif",
           "src/test/resources/webgoat_v8.2.2_codeql.sarif",
-          "--codemod-exclude=pixee:java/resource-leak",
           "--dont-exit",
           repoDir.getPath()
         });
--- a/core-codemods/src/intTest/java/io/codemodder/integration/WebGoat822Test.java
+++ b/core-codemods/src/intTest/java/io/codemodder/integration/WebGoat822Test.java
@@ -95,7 +95,7 @@ final class WebGoat822Test extends GitRepositoryTest {
             .flatMap(Collection::stream)
             .collect(Collectors.toList());
 
-    assertThat(fileChanges.size(), is(67));
+    assertThat(fileChanges.size(), is(71));
 
     // we only inject into a couple files
     verifyStandardCodemodResults(fileChanges);
--- a/core-codemods/src/intTest/java/io/codemodder/integration/WebGoat822Test.java
+++ b/core-codemods/src/intTest/java/io/codemodder/integration/WebGoat822Test.java
@@ -95,7 +95,7 @@ final class WebGoat822Test extends GitRepositoryTest {
             .flatMap(Collection::stream)
             .collect(Collectors.toList());
 
-    assertThat(fileChanges.size(), is(58));
+    assertThat(fileChanges.size(), is(67));
 
     // we only inject into a couple files
     verifyStandardCodemodResults(fileChanges);
@@ -117,6 +117,7 @@ final class WebGoat822Test extends GitRepositoryTest {
           outputFile.getPath(),
           "--sarif",
           "src/test/resources/webgoat_v8.2.2_codeql.sarif",
+          "--codemod-exclude=pixee:java/resource-leak",
           "--dont-exit",
           repoDir.getPath()
         });
--- a/core-codemods/src/intTest/java/io/codemodder/integration/WebGoat822Test.java
+++ b/core-codemods/src/intTest/java/io/codemodder/integration/WebGoat822Test.java
@@ -95,7 +95,7 @@ final class WebGoat822Test extends GitRepositoryTest {
             .flatMap(Collection::stream)
             .collect(Collectors.toList());
 
-    assertThat(fileChanges.size(), is(71));
+    assertThat(fileChanges.size(), is(58));
 
     // we only inject into a couple files
     verifyStandardCodemodResults(fileChanges);
--- a/core-codemods/src/main/java/io/codemodder/codemods/ResourceLeakCodemod.java
+++ b/core-codemods/src/main/java/io/codemodder/codemods/ResourceLeakCodemod.java
@@ -13,7 +13,7 @@ import java.util.stream.Collectors;
     id = "pixee:java/resource-leak",
     reviewGuidance = ReviewGuidance.MERGE_WITHOUT_REVIEW,
     importance = Importance.MEDIUM,
-    executionPriority = CodemodExecutionPriority.HIGH)
+    executionPriority = CodemodExecutionPriority.LOW)
 public final class ResourceLeakCodemod extends JavaParserChanger {
 
   private Optional<CodemodChange> onNodeFound(final Expression expr) {
--- a/core-codemods/src/main/java/io/codemodder/codemods/ResourceLeakCodemod.java
+++ b/core-codemods/src/main/java/io/codemodder/codemods/ResourceLeakCodemod.java
@@ -1,0 +1,37 @@
+package io.codemodder.codemods;
+
+import com.github.javaparser.ast.CompilationUnit;
+import com.github.javaparser.ast.expr.Expression;
+import io.codemodder.*;
+import io.codemodder.javaparser.JavaParserChanger;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+/** A codemod that wraps AutoCloseable objects whenever possible. */
+@Codemod(
+    id = "pixee:java/resource-leak",
+    reviewGuidance = ReviewGuidance.MERGE_WITHOUT_REVIEW,
+    importance = Importance.MEDIUM,
+    executionPriority = CodemodExecutionPriority.HIGH)
+public final class ResourceLeakCodemod extends JavaParserChanger {
+
+  private Optional<CodemodChange> onNodeFound(final Expression expr) {
+    int originalLine = expr.getBegin().get().line;
+    if (ResourceLeakFixer.checkAndFix(expr).isPresent()) {
+      return Optional.of(CodemodChange.from(originalLine));
+    } else {
+      return Optional.empty();
+    }
+  }
+
+  @Override
+  public CodemodFileScanningResult visit(
+      final CodemodInvocationContext context, final CompilationUnit cu) {
+    List<CodemodChange> changes =
+        cu.findAll(Expression.class).stream()
+            .flatMap(expr -> onNodeFound(expr).stream())
+            .collect(Collectors.toList());
+    return CodemodFileScanningResult.withOnlyChanges(changes);
+  }
+}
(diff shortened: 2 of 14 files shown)


core-codemods/src/intTest/java/io/codemodder/integration/WebGoat822Test.java
```diff
@@ -117,6 +117,7 @@ void it_transforms_webgoat_with_codeql() throws Exception {
           outputFile.getPath(),
           "--sarif",
           "src/test/resources/webgoat_v8.2.2_codeql.sarif",
```
nahsra, line 119: This could be solved by making the CodeQL one run with `CodemodExecutionPriority.LOW`, maybe?

core-codemods/src/main/java/io/codemodder/codemods/ResourceLeakCodemod.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+/** A codemod that wraps AutoCloseable objects whenever possible. */
+@Codemod(
+    id = "pixee:java/resource-leak",
+    reviewGuidance = ReviewGuidance.MERGE_WITHOUT_REVIEW,
+    importance = Importance.MEDIUM,
+    executionPriority = CodemodExecutionPriority.HIGH)
+public final class ResourceLeakCodemod extends JavaParserChanger {
```
nahsra, line 17: I think this should be `CodemodExecutionPriority.LOW`.

core-codemods/src/main/java/io/codemodder/codemods/ResourceLeakFixer.java
```diff
@@ -53,15 +54,6 @@ public static Optional<Integer> checkAndFix(final Expression expr) {
 
   /**
    * Detects if a {@link Expression} detected by CodeQL that creates a leaking resource is fixable.
```
nahsra, line 56: Did these assumptions change, because we're not using a more general utility that doesn't guarantee the same things by this point?

core-codemods/src/main/java/io/codemodder/codemods/ResourceLeakFixer.java
```diff
@@ -97,16 +98,44 @@ public static boolean isFixable(final Expression expr) {
...
+      return true;
+    }
+    // find all the variables it flows into
+    final var allVariables = flowsInto(expr);
+
+    // flows into anything that is not a local variable or parameter
+    if (allVariables.stream().anyMatch(Either::isRight)) {
+      return true;
+    }
+
```
nahsra, line 125: I love how easy it is to express these constraints now.

core-codemods/src/test/resources/resource-leak/Test.java.before
```diff
@@ -0,0 +1,52 @@
...
+  public void nestedReaders() throws IOException {
+    BufferedReader br = new BufferedReader(new BufferedReader(new FileReader("~/test.txt")));
+  }
+
+  public Reader noFixReturned() throws IOException {
+    BufferedReader br = new BufferedReader(new FileReader(""));
+    BufferedReader br2 = new BufferedReader(br);
+    return br2;
+  }
+
```
nahsra, line 40: Can we call this `noFixMethod` like the others so we know the intention of the test code?

core-codemods/src/test/resources/resource-leak/Test.java.after
```diff
@@ -0,0 +1,58 @@
...
+    }
+  }
+
+  public void multipleResourcesInOneLine(String query, ResultSet rs) throws SQLException {
+    try (var stream = conn.createStatement().executeQuery(query)) {
+    }
+  }
+
+  public void nestedReaders() throws IOException {
+    try (FileReader resource1 = new FileReader("~/test.txt");
```
nahsra, line 35: I think `resource1` is fine for now, but I think we need to create a follow-up ticket for calling it `fileReader` (whatever the type is, at least).

core-codemods/src/test/resources/prevent-filewriter-leak-with-nio/Test.java.after
```diff
@@ -9,18 +9,14 @@ public class Test
 {
     private Response response;
     void foo(File f) {
-        // ruleid: prevent-filewriter-leak-with-nio
-        Writer change1 = Files.newBufferedWriter(f.toPath());
-        // ruleid: prevent-filewriter-leak-with-nio
-        Writer change2 = Files.newBufferedWriter(f.toPath());
```
nahsra, line 15: This test makes me wonder if we should have try-with-resources with multiple files. Consider the example [here](https://stackoverflow.com/questions/47175526/try-with-multiple-resource-in-java):

```
try (
        java.util.zip.ZipFile zf =
             new java.util.zip.ZipFile(zipFileName);
        java.io.BufferedWriter writer = 
            java.nio.file.Files.newBufferedWriter(outputFilePath, charset)
    ) {
        // Enumerate each entry
        for (java.util.Enumeration entries =
                                zf.entries(); entries.hasMoreElements();) {
            // Get the entry name and write it to the output file
            String newLine = System.getProperty("line.separator");
            String zipEntryName =
                 ((java.util.zip.ZipEntry)entries.nextElement()).getName() +
                 newLine;
            writer.write(zipEntryName, 0, zipEntryName.length());
        }
    }
```

Is this worth chasing down? Should it be added to the scope of this change, or added later?

nahsra (changes requested):
How do you like this plan:
* Take the new codemod out of `DefaultCodemods`
* Create tickets for the various improvements discussed here
* Ship this PR 
* Work follow-on tickets
* Add back to `DefaultCodemods`

That way, we get to keep moving forward, improve incrementally, and don't have to give our users changes they may have formatting objections towards.

nahsra:
Also, this PR and the changes are _awesome_!

core-codemods/src/test/resources/prevent-filewriter-leak-with-nio/Test.java.after
andrecsilva: You mean having a try with multiple resources instead of two consecutive ones? I can work on a transform to do that.

core-codemods/src/main/java/io/codemodder/codemods/ResourceLeakFixer.java
andrecsilva: (1) and (3) are being tested by the new `isClosedOrEscapes`. This test was already in place for dependent resources, so I just added it to the resource we are trying to fix.
(2) Is not being tested. This decision was deliberate. This means that, for example, if you have a `Connection` object that is already being closed, it will still try to close a `Statement` object that it spawns. The reason being that most examples and advice I've encountered recommended to close those object separately.
Generally speaking, this won't cause any trouble since `close()` is usually idempotent (it is not required to, but strongly encouraged, see https://docs.oracle.com/javase/8/docs/api/java/lang/AutoCloseable.html). I've yet to test one that is not on the JAVA API. I can also include a test for this property if you want to be on the safe side.

core-codemods/src/main/java/io/codemodder/codemods/ResourceLeakCodemod.java
nahsra: Let me defend why. I say `LOW` because I'd like CodeQL to fix it "first", then I'd love any codemods that want to orchestrate this ability into themselves be able to do it next, then this as a last resort.

core-codemods/src/main/java/io/codemodder/codemods/ResourceLeakCodemod.java
```diff
@@ -0,0 +1,37 @@
...
+    if (ResourceLeakFixer.checkAndFix(expr).isPresent()) {
+      return Optional.of(CodemodChange.from(originalLine));
+    } else {
+      return Optional.empty();
+    }
+  }
+
+  @Override
+  public CodemodFileScanningResult visit(
+      final CodemodInvocationContext context, final CompilationUnit cu) {
```
nahsra, line 30: This will be an extremely expensive codemod to execute as it will force every file to be visited, and every `Expression` to be evaluated. 

I think we need a Semgrep pattern to detect reasonable starting points that can be confirmed more thoroughly by AST-inspecting code.

core-codemods/src/main/java/io/codemodder/codemods/ResourceLeakFixer.java
```diff
@@ -285,40 +321,40 @@ private static Either<LocalDeclaration, Node> isLocalDeclaration(Node n) {
     return Either.right(n);
   }
 
+  /** Checks if {@code expr} creates an AutoCloseable Resource. */
+  private static boolean isFilesResourceInit(final MethodCallExpr expr) {
+    try {
+      var hasFilesScope =
+          expr.getScope()
+              .map(mce -> mce.calculateResolvedType().describe())
+              .filter("java.nio.file.Files"::equals);
```
nahsra, line 330: The `startsWith("new")` makes me a bit nervous here as it doesn't feel future safe.

core-codemods/src/main/java/io/codemodder/codemods/ResourceLeakCodemod.java
nahsra: This is a guess, but I think it's not that much to test. Again, maybe we should add it to another ticket, and make sure we check that ticket off before we add to `DefaultCodemods`.

core-codemods/src/main/java/io/codemodder/codemods/ResourceLeakFixer.java
nahsra: I think this is right.

core-codemods/src/main/java/io/codemodder/codemods/ResourceLeakCodemod.java
andrecsilva: The only reason it visits `Expression` is that we cannot express sum types easily. Notice that the first thing it does is test if the given expression is `ObjectCreationExpression` or `MethodCallExpression`. After that the first check is if it is a `AutoCloseable` type. I think those two check are fast and restrictive enough to nip most visits without spending much.

core-codemods/src/main/java/io/codemodder/codemods/ResourceLeakFixer.java
andrecsilva: True enough, I will change it to check against a list of current resource creation methods in `Files`.

nahsra (approved):

Issue #314, opened 2024-03-20
Populate `detectionTool` metadata for Semgrep codemods


drdavella:

nahsra:
Given the changes just added in https://github.com/pixee/codemodder-java/pull/335, let's add this same metadata to Semgrep codemod changes.

PR #343, opened 2024-04-09
Populate detectionTool metadata for Semgrep codemods

carlosu7:
Populate detectionTool metadata for Semgrep codemods  (`@ProvidedSemgrepScan` and `@SemgrepScan`)

Issue https://github.com/pixee/codemodder-java/issues/314


nahsra (changes requested):
The purpose of the ticket was to only enhance codemods that fix things provided by an external Semgrep scan. Also, I think it would make sense to have a `SemgrepSarifJavaParserChanger` that provided the `vendorName()` so we won't have a lot of duplicate code.

PR #347, opened 2024-04-11
Populate detectionTool metadata for CodeQL codemods

carlosu7:
Implement `FixOnlyCodeChanger` for CodeQL codemods


core-codemods/src/main/java/io/codemodder/codemods/MavenSecureURLCodemod.java
```diff
@@ -70,7 +90,9 @@ private CodemodFileScanningResult processXml(final Path file)
     Set<Integer> linesAffected = xmlChange.linesAffected();
 
     List<CodemodChange> allWeaves =
-        linesAffected.stream().map(CodemodChange::from).collect(Collectors.toList());
+        linesAffected.stream()
```
nahsra, line 93: Is `file.toString()` the ID? Shouldn't we be getting a GUID out of the SARIF `Result` object or something? This ID represents the tool vendor's ID for that finding.

plugins/codemodder-plugin-codeql/src/main/java/io/codemodder/providers/sarif/codeql/CodeQLSarifJavaParserChanger.java
```diff
@@ -0,0 +1,26 @@
+package io.codemodder.providers.sarif.codeql;
+
+import com.contrastsecurity.sarif.Result;
+import com.github.javaparser.ast.Node;
+import io.codemodder.RuleSarif;
+import io.codemodder.SarifPluginFixOnlyCodeChanger;
+import io.codemodder.SourceCodeRegionExtractor;
+
+/**
+ * Provides foundational functionality for modifying Java code using JavaParser based on findings
```
nahsra, line 10: Comment copied and incorrect

nahsra (changes requested):

core-codemods/src/main/java/io/codemodder/codemods/MavenSecureURLCodemod.java
carlosu7: @nahsra currently, i'm not seeing how the results are related to the affected lines of the xmlChange because the list of results right now is not taken in consideration and i think it should now be considered if we want to link the result guid to the affected line

but since the current test only has 1 affected line and the result guid is null, i'm not sure how to link the guid and the affected line

core-codemods/src/main/java/io/codemodder/codemods/MavenSecureURLCodemod.java
```diff
@@ -29,7 +30,8 @@
     reviewGuidance = ReviewGuidance.MERGE_WITHOUT_REVIEW,
     importance = Importance.MEDIUM,
     executionPriority = CodemodExecutionPriority.HIGH)
-public final class MavenSecureURLCodemod extends SarifPluginRawFileChanger {
+public final class MavenSecureURLCodemod extends SarifPluginRawFileChanger
```
nahsra, line 33: Should this extend a new `CodeQLSarifPluginRawFileChanger` type?

core-codemods/src/main/java/io/codemodder/codemods/MavenSecureURLCodemod.java
nahsra: Is the correlation GUID also null? If not, we should use that.

plugins/codemodder-plugin-codeql/src/main/java/io/codemodder/providers/sarif/codeql/CodeQLSarifJavaParserChanger.java
```diff
@@ -0,0 +1,26 @@
...
+ * from a SARIF file generated by Semgrep analysis.
+ */
```
nahsra, line 12: Still a typo here, it says "Semgrep"...

nahsra (changes requested):

PR #351, opened 2024-04-16
Added workaround for children out of order while adding statements.

### LINKED ISSUE #340: Add RemoveUnusedLocalVariables transform to SQLParameterizerWithCleanup
The sql parameterizer codemod may leave a few unused variables after the transformation. This transformation already exists in `ASTTransforms.removeUnusedLocalVariables` but it currently fails due to a bug in JavaParser where children may appear out of order after adding a statement.

This can be seen by executing the following snippet:
```java
   String code =
        """
    class A{
	    void f(){
		    int b = 2;
		    int c = 3;
	    }
    }
    """;
    CompilationUnit cu = StaticJavaParser.parse(code);
    System.out.println(cu);
    // var name = cu.findAll(NameExpr.class).get(1);
    var stmt = cu.findAll(Statement.class).get(1);
    var newStmt = StaticJavaParser.parseStatement("int a = 1;");
    var blockStmt = (BlockStmt) stmt.getParentNode().get();
    System.out.println("Children: " + blockStmt.getChildNodes());
    System.out.println("Statements: " + blockStmt.getStatements());
    blockStmt.addStatement(0, newStmt);
    System.out.println(cu);
    System.out.println(stmt);
    System.out.println("Children: " + blockStmt.getChildNodes());
    System.out.println("Statements: " + blockStmt.getStatements());
```

this will produce:
```
    class A {

        void f() {
            int b = 2;
            int c = 3;
        }
    }

    Children: [int b = 2;, int c = 3;]
    Statements: [int b = 2;, int c = 3;]
    class A {

        void f() {
            int a = 1;
            int b = 2;
            int c = 3;
        }
    }

    int b = 2;
    Children: [int b = 2;, int c = 3;, int a = 1;]
    Statements: [int a = 1;, int b = 2;, int c = 3;]
```

Notice that the children and statement orders differ after the addition.

andrecsilva:
- Added a new method to add statements to nodes that preserves children order. This unfortunately causes some small formatting changes.
- Enabled cleanup step that removes unused variables for some parameterization codemods.
Closes #340.

### DIFF
--- a/core-codemods/src/test/resources/hql-parameterizer/Test.java.before
+++ b/core-codemods/src/test/resources/hql-parameterizer/Test.java.before
@@ -9,16 +9,19 @@ public final class Test {
 
     void direct(String tainted) {
       Query<User> hqlQuery = session.createQuery("select p from Person p where p.name like '" + tainted + "'");
+      List l = hqlQuery.list();
     }
 
     void indirect(String tainted) {
       String query = "select p from Person p where p.name like '" + tainted + "'";
       Query<User> hqlQuery = session.createQuery(query);
+      List l = hqlQuery.list();
     }
 
     void indirectMultiple(String tainted, String tainted2) {
       String query = "select p from Person p where p.name like '" + tainted + "' and p.surname like '" + tainted2 + "'";
       Query<User> hqlQuery = session.createQuery(query);
+      List l = hqlQuery.list();
     }
 
     void indirectMultipeString(String tainted, String tainted2) {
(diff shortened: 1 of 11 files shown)


core-codemods/src/test/resources/hql-parameterizer/Test.java.before
```diff
@@ -9,22 +9,26 @@ public final class Test {
 
     void direct(String tainted) {
       Query<User> hqlQuery = session.createQuery("select p from Person p where p.name like '" + tainted + "'");
+      List l = hqlQuery.list();
     }
 
     void indirect(String tainted) {
       String query = "select p from Person p where p.name like '" + tainted + "'";
       Query<User> hqlQuery = session.createQuery(query);
```
nahsra, line 17: Why are we adding these `hqlQuery.list()` calls?

andrecsilva: To avoid being removed as unused by the cleanup step.

nahsra (approved):

PR #352, opened 2024-04-16
✅ RawFileChangerTest Allows Tests to Specify "No Changes" Case

gilday:
The `LLMVerifyingCodemodTestMixin` and `CodemodTestMixin` types allow the test author to specify cases where the codemod should not act on a file. This change adds the same capability to the `RawFileChangerTest` fixture.

Also, this enhances the test fixture to detect the case where the user has improperly set-up their test.

/towards ISS-834

### DIFF
--- a/framework/codemodder-testutils/src/main/java/io/codemodder/testutils/RawFileCodemodTest.java
+++ b/framework/codemodder-testutils/src/main/java/io/codemodder/testutils/RawFileCodemodTest.java
@@ -93,7 +93,12 @@ public interface RawFileCodemodTest {
     assertThat(result.getReferences(), is(not(empty())));
 
     final var modifiedFile = Files.readString(tmpFilePath);
-    assertThat(modifiedFile, equalTo(Files.readString(filePathAfter)));
+    if (filePathAfter == null) {
+      // lack of an after file indicates that the before file should be unchanged
+      assertThat(modifiedFile, equalTo(Files.readString(filePathBefore)));
+    } else {
+      assertThat(modifiedFile, equalTo(Files.readString(filePathAfter)));
+    }
     Files.deleteIfExists(tmpFilePath);
   }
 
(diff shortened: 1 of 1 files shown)


framework/codemodder-testutils/src/main/java/io/codemodder/testutils/RawFileCodemodTest.java
```diff
@@ -110,23 +115,31 @@ private void verifyCodemod(
...
+    final Map<String, Path> afterFilesMap;
+    try (final var files = Files.list(testResourceDir)) {
+      afterFilesMap =
+          files
+              .filter(file -> file.getFileName().toString().endsWith(".after"))
+              .collect(Collectors.toMap(RawFileCodemodTest::trimExtension, f -> f));
+    }
+    if (allBeforeFiles.isEmpty()) {
+      throw new IllegalArgumentException("No .before files found in " + testResourceDir);
+    }
```
gilday, line 142: Added this check, because I was stuck trying to figure out why my test was passing when I expected it to fail, and it had to do with an incorrect directory structure in `src/test/resources`.

framework/codemodder-testutils/src/main/java/io/codemodder/testutils/RawFileCodemodTest.java
```diff
@@ -110,23 +115,31 @@ private void verifyCodemod(
...
+      allBeforeFiles =
+          files.filter(file -> file.getFileName().toString().endsWith(".before")).toList();
+    }
+    final Map<String, Path> afterFilesMap;
+    try (final var files = Files.list(testResourceDir)) {
+      afterFilesMap =
+          files
+              .filter(file -> file.getFileName().toString().endsWith(".after"))
+              .collect(Collectors.toMap(RawFileCodemodTest::trimExtension, f -> f));
+    }
```
gilday, line 139: `Files.list` must be closed to avoid leaks.

nahsra (approved):

Issue #359, opened 2024-04-19
Expected fix lines mapping doesn't work with multi-file test feature


nahsra:
There is a bug in the test framework that makes `expectedFixesAtLines` fail when you have multiple test files. 

There is one set of line numbers to look for, but the test `@Metadata` allows you to specify multiple test files. Therefore, for each file tested, the framework will looks for all of the numbers. We have to figure out a way to map a set of expected findings to each individual test file, or throw an error when there are multiple files, because this was painful to debug.

PR #361, opened 2024-04-22
Expected fix lines mapping doesn't work with multi-file test feature

carlosu7:
Throw an error when there are multiple files and expected fix lines. 

Issue https://github.com/pixee/codemodder-java/issues/359


framework/codemodder-testutils/src/main/java/io/codemodder/testutils/CodemodTestMixin.java
```diff
@@ -40,6 +40,13 @@ default Stream<DynamicTest> generateTestCases(@TempDir final Path tmpDir) throws
             .filter(Files::isRegularFile)
             .filter(path -> path.toString().endsWith(".java.before"));
```
nahsra, line 42: Can we offer a test for this?

Also, for the MavenSecureURL codemod, can we break the test into multiple tests for each different directory, so we can test the expected file fixe lines? That way we'll be fixing the original issue.

nahsra (changes requested):

framework/codemodder-testutils/src/main/java/io/codemodder/testutils/MetadataUtil.java
```diff
@@ -0,0 +1,32 @@
+package io.codemodder.testutils;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+
```
nahsra, line 7: There's already a file called `ExpectedFixes` that seems like we should be adding this towards rather than creating a new utility or pattern.

Also, incidentally, the `*Util` naming pattern has been superseded generally in Java (and specifically also in this project) with a pattern like `Objects`, `ExpectedFixes`, etc.

nahsra (changes requested):

framework/codemodder-testutils/src/test/java/io/codemodder/testutils/ExpectedFixesTest.java
```diff
@@ -0,0 +1,77 @@
...
+  void it_expects_no_exception_fail_fix_line_single_before_file() throws IOException {
+
+    try {
+      ExpectedFixes.checkExpectedFixLinesUsage(SINGLE_BEFORE_FILE_DIR, new int[] {}, new int[] {1});
+    } catch (final IllegalArgumentException ex) {
+      fail("Exception not expected");
+    }
+  }
+
+  @Test
```
nahsra, line 68: I would guess these tests could all just exist without the try/catch block. was there a reason we added them?

framework/codemodder-testutils/src/main/java/io/codemodder/testutils/ExpectedFixes.java
```diff
@@ -18,10 +21,15 @@ private ExpectedFixes() {}
 
   /** Verify the expected fix metadata. */
   static void verifyExpectedFixes(
+      final Path testResourceDir,
       final CodeTFResult result,
       final CodeChanger changer,
       final int[] expectedFixLines,
-      final int[] expectingFailedFixesAtLines) {
+      final int[] expectingFailedFixesAtLines)
+      throws IOException {
```
nahsra, line 29: Approved, but probably no reason to reference this method statically? Can just call `checkExpectedFixLinesUsage()` directly?

Issue #364, opened 2024-05-06
Handle comma-separated filename inputs for Sonar issues JSON and Sonar hotspots JSON


drdavella:
The `--sarif` flag currently handles comma-separated lists of files as inputs. We need to do the same for `--sonar-issues-json` and `--sonar-hotspots-json`.

carlosu7:
@drdavella i will close this issue

PR #365, opened 2024-05-07
Handle comma-separated filename inputs for Sonar issues JSON and Sonar hotspots JSON

carlosu7:
Issue: https://github.com/pixee/codemodder-java/issues/364

Example of running a list of sonar findings files:
```

./gradlew :core-codemods:run --args='--output /Users/iwa/Documents/GitHub/WebGoat_12_23/WebGoat_12_23.codetf --verbose --sonar-issues-json /Users/iwa/Desktop/sonar1.json,/Users/iwa/Desktop/sonar2.json --log-format human /Users/iwa/Documents/GitHub/WebGoat_12_23/'
```

Sonar hotspot CLI input was the only update since there's no implementation for it yet


framework/codemodder-base/src/main/java/io/codemodder/CodemodLoader.java
```diff
@@ -28,7 +28,7 @@ public CodemodLoader(
       final List<Path> includedFiles,
       final Map<String, List<RuleSarif>> ruleSarifByTool,
       final List<ParameterArgument> codemodParameters,
-      final Path sonarIssuesJsonFile,
+      final List<Path> sonarIssuesJsonFile,
```
drdavella, line 31: Change variable name here and below
```suggestion
      final List<Path> sonarIssuesJsonFiles,
```

framework/codemodder-base/src/main/java/io/codemodder/CLI.java
```diff
@@ -378,8 +379,11 @@ public Integer call() throws IOException {
 
       // create the loader
       CodeDirectory codeDirectory = new DefaultCodeDirectory(projectPath);
-      List<Path> sarifFiles =
-          sarifs != null ? sarifs.stream().map(Path::of).collect(Collectors.toList()) : List.of();
+      List<Path> sarifFiles = sarifs != null ? sarifs.stream().map(Path::of).toList() : List.of();
```
drdavella, line 382: Could this be factored out into a function to share?

drdavella (changes requested):
Looks good overall. Even though `--sonar-hotspots-json` is just a pass-thru without implementation I think we should also update the help message to indicate a comma-separated input.

plugins/codemodder-plugin-sonar/src/main/java/io/codemodder/providers/sonar/SonarModule.java
```diff
@@ -21,32 +21,39 @@ final class SonarModule extends AbstractModule {
...
-      try {
-        issueResponse =
-            new ObjectMapper()
-                .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
-                .readValue(issuesFile.toFile(), SearchIssueResponse.class);
-        allIssues = issueResponse.getIssues();
-      } catch (IOException e) {
-        throw new UncheckedIOException("Problem reading Sonar issues JSON file", e);
-      }
+    if (issuesFiles != null) {
```
drdavella, line 39: This code makes sense to me but I also want to be careful that we don't introduce any regressions here. Would you be able to add a unit test or do you feel like this is sufficiently covered by existing tests?

carlosu7: @drdavella 
When we run a codemod test that is not a sonar codemod, this sonar module is being executed because all modules (defectdojo, semgrep, codeql, sonar, etc) are being configured at CodemodLoader.java. So, in when the non-sonar codemod test is being executed, this sonar module handles the empty/null issuesFiles because they're not required for that codemod (defectdojo, semgrep, etc).

So yes, this case is already covered.

carlosu7: @drdavella In this commit https://github.com/pixee/codemodder-java/pull/365/commits/be04cc0fdfc8ed7ec6ddb763d765e804ac8da868 I have updated our tests to support multiple sonar issues json files and refactor one existing sonar test

framework/codemodder-testutils/src/main/java/io/codemodder/testutils/CodemodTestMixin.java
```diff
@@ -111,8 +114,21 @@ private void verifyCodemod(
...
+        sonarIssuesJsonFiles != null ? Arrays.asList(sonarIssuesJsonFiles) : new ArrayList<>();
+
+    final List<Path> sonarJsonsPaths =
+        sonarJsons.stream()
+            .map(testResourceDir::resolve)
+            .filter(Files::exists)
+            .collect(Collectors.toList());
+
+    if (sonarJsonsPaths.isEmpty()) {
+      Path defaultPath = testResourceDir.resolve("sonar-issues.json");
```
drdavella, line 127: Sorry, I thought I left this comment before but I don't see it. Shouldn't it be this now?
```suggestion
      Path defaultPath = testResourceDir.resolve("sonar-issues_1.json");
```

carlosu7: @drdavella no, default sonar issues filename remains the same as it previously was, otherwise, i will have to rename all other sonar codemod test json files 
Screenshot 2024-05-13 at 12 51 17 p m

drdavella: Oh okay, I'm sorry. I missed that the one you renamed was for a specific test.

carlosu7: yes

PR #377, opened 2024-06-11
Disable ReplaceDefaultHttpClientCodemodIntegrationTest

carlosu7:
Test is failing dua connection timeout

You can see workflow report test failure at https://github.com/pixee/codemodder-java/actions/runs/9470753410?pr=373


nahsra (approved):

nahsra:
Worth noting here for posterity that both the original and the transformed code in this integration test are failing, and both the other integration tests are passing -- so there's something specific to that test or container that is problematic. It has probably been broken for weeks and it will take some time to get to the root cause.

PR #378, opened 2024-06-11
Modifications to read both types of sonar findings: issues and hotspots

carlosu7:
Read data using issues and hotspots parent abstract class: `SonarFinding`.

This change will allow us to start creating sonar hotspots codemods


core-codemods/src/main/java/io/codemodder/codemods/AddMissingOverrideCodemod.java
```diff
@@ -24,16 +25,17 @@ public final class AddMissingOverrideCodemod extends SonarPluginJavaParserChange
 
   @Inject
   public AddMissingOverrideCodemod(
-      @ProvidedSonarScan(ruleId = "java:S1161") final RuleIssues issues) {
+      @ProvidedSonarScan(type = SonarFindingType.ISSUE, ruleId = "java:S1161")
```
nahsra, line 28: These ergonomics don't feel right. Why can't we support both of these:

`@ProvidedSonarScan(ruleId = "java:S1161") final Hotspots hotspots)`
..
`@ProvidedSonarScan(ruleId = "java:S1161") final Issues issues)`

It does seem like Guice can support both of these?

nahsra (changes requested):
I don't want to have to downcast in each codemod to get `Hotspot`-specific or `Issue`-specific fields. Please create Guice bindings for each subclass and take away the `type` annotation parameter.

core-codemods/src/main/java/io/codemodder/codemods/AvoidImplicitPublicConstructorCodemod.java
```diff
@@ -38,7 +38,7 @@ public ChangesResult onIssueFound(
       final CodemodInvocationContext context,
       final CompilationUnit cu,
       final SimpleName simpleName,
```
nahsra, line 40: This is weird, shouldn't this be variable be 'issue' still?

core-codemods/src/main/java/io/codemodder/codemods/FixRedundantStaticOnEnumCodemod.java
```diff
@@ -18,11 +18,11 @@
     importance = Importance.LOW,
     executionPriority = CodemodExecutionPriority.HIGH)
 public final class FixRedundantStaticOnEnumCodemod
-    extends SonarPluginJavaParserChanger<EnumDeclaration> {
+    extends SonarIssuesPluginJavaParserChanger<EnumDeclaration> {
```
nahsra, line 21: Have you considered making this generic instead, that way we wouldn't need subtypes?

nahsra (changes requested):
Let's make sure the variables are named right, and consider generifying the Sonar JavaParserChanger so we don't have to create new subclasses for each Sonar issue type and repeat a lot of code.

plugins/codemodder-plugin-sonar/src/main/java/io/codemodder/providers/sonar/RuleFinding.java
```diff
@@ -0,0 +1,15 @@
+package io.codemodder.providers.sonar;
+
+import io.codemodder.sonar.model.SonarFinding;
+import java.nio.file.Path;
+import java.util.List;
+
+/** A view of the Sonar findings results file for a given rule. */
```
nahsra, line 7: This seems like this is another opportunity to be genercized.

```
interface RuleFinding<T extends SonarFinding>
...
  List<T> getResultsByPath(Path path);
```

plugins/codemodder-plugin-sonar/src/main/java/io/codemodder/providers/sonar/RuleHotspot.java
```diff
@@ -0,0 +1,27 @@
+package io.codemodder.providers.sonar;
+
+import io.codemodder.sonar.model.Hotspot;
+import io.codemodder.sonar.model.SonarFinding;
+import java.nio.file.Path;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
```
nahsra, line 9: All `public` types should have JavaDoc.

plugins/codemodder-plugin-sonar/src/main/java/io/codemodder/providers/sonar/RuleHotspot.java
```diff
@@ -0,0 +1,27 @@
...
+public final class RuleHotspot implements RuleFinding {
+
+  private final Map<String, List<SonarFinding>> findings;
+
```
nahsra, line 13: I think there's room for more optimization of this class design.

It seems like `RuleHotspot` should be an interface like...
```java
public interface Hotspot implements RuleFinding<SonarHotspot> {
  static Hotspot from(Map<String,List<SonarHotspot>> hotspots) {
      return new DefaultHotspot(hotspots);
  }
}
```

We know which are hotspots, and which are issues, when we parse them. Then, when it gets to this point, we have to check and filter by casting. We should figure this out once, and never have to figure it out again.

plugins/codemodder-plugin-sonar/src/main/java/io/codemodder/providers/sonar/RuleIssue.java
```diff
@@ -0,0 +1,27 @@
+package io.codemodder.providers.sonar;
+
+import io.codemodder.sonar.model.Issue;
+import io.codemodder.sonar.model.SonarFinding;
+import java.nio.file.Path;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
```
nahsra, line 9: JavaDoc. And, this type seems identical to and should share some of the logic from the similar type for hotspots.

nahsra (changes requested):

nahsra (approved):

PR #383, opened 2024-06-14
SQL Injection visitor utility

carlosu7:
Utility to help implement sql injection codemods' visit method

### DIFF
--- a/core-codemods/src/main/java/io/codemodder/codemods/util/DefaultJavaParserSQLInjectionRemediatorStrategy.java
+++ b/core-codemods/src/main/java/io/codemodder/codemods/util/DefaultJavaParserSQLInjectionRemediatorStrategy.java
@@ -16,47 +16,58 @@ import java.util.List;
 import java.util.function.Function;
 
 /** Utility class for visiting and processing potential SQL injection points in Java source code. */
-public class SqlInjectionVisitor {
+public class DefaultJavaParserSQLInjectionRemediatorStrategy
+    implements JavaParserSQLInjectionRemediatorStrategy {
 
-  private SqlInjectionVisitor() {}
-
-  public static <T> CodemodFileScanningResult visit(
+  /**
+   * Visits the provided CompilationUnit and processes findings for potential SQL injections.
+   *
+   * @param context the context of the codemod invocation
+   * @param cu the compilation unit to be scanned
+   * @param pathFindings a collection of findings to be processed
+   * @param detectorRule the rule used to detect potential issues
+   * @param findingIdExtractor a function to extract the ID from a finding
+   * @param findingLineExtractor a function to extract the line number from a finding
+   * @param <T> the type of the findings
+   * @return a result object containing the changes and unfixed findings
+   */
+  public <T> CodemodFileScanningResult visit(
       final CodemodInvocationContext context,
       final CompilationUnit cu,
-      Collection<T> pathFindings,
+      final Collection<T> pathFindings,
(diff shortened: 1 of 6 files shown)


core-codemods/src/main/java/io/codemodder/codemods/util/SqlInjectionVisitor.java
```diff
@@ -0,0 +1,99 @@
...
+import io.codemodder.codemods.SQLParameterizerWithCleanup;
+import io.codemodder.codetf.DetectorRule;
+import io.codemodder.codetf.FixedFinding;
+import io.codemodder.codetf.UnfixedFinding;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.function.Function;
+
+/** Utility class for visiting and processing potential SQL injection points in Java source code. */
```
nahsra, line 18: I think this should be an `interface` with a default implementation you can create (or reference as an instance field). E.g.:

```
interface JavaParserSQLInjectionRemediatorStrategy {
  (methods)

  JavaParserSQLInjectionRemediatorStrategy DEFAULT  = new DefaultJavaParserSQLInjectionRemediatorStrategy();
```

This way we can inject it as a seam into all these different tool types.

core-codemods/src/main/java/io/codemodder/codemods/util/SqlInjectionVisitor.java
```diff
@@ -0,0 +1,99 @@
...
+public class SqlInjectionVisitor {
+
+  private SqlInjectionVisitor() {}
+
```
nahsra, line 22: Some of these are final, and there's no JavaDoc

A method with this many arguments is a code smell to me. Not sure what advice give on it at the moment

nahsra (changes requested):

core-codemods/src/main/java/io/codemodder/codemods/util/DefaultJavaParserSQLInjectionRemediatorStrategy.java
```diff
@@ -0,0 +1,110 @@
...
+import io.codemodder.codetf.DetectorRule;
+import io.codemodder.codetf.FixedFinding;
+import io.codemodder.codetf.UnfixedFinding;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.function.Function;
+
+/** Utility class for visiting and processing potential SQL injection points in Java source code. */
+public class DefaultJavaParserSQLInjectionRemediatorStrategy
```
nahsra, line 19: This should be `final` and probably doesn't need to be `public`. Let's just store a reference to it in the already-public `interface`.

core-codemods/src/main/java/io/codemodder/codemods/util/JavaParserSQLInjectionRemediatorStrategy.java
```diff
@@ -0,0 +1,34 @@
+package io.codemodder.codemods.util;
+
+import com.github.javaparser.ast.CompilationUnit;
+import io.codemodder.CodemodFileScanningResult;
+import io.codemodder.CodemodInvocationContext;
+import io.codemodder.codetf.DetectorRule;
+import java.util.Collection;
+import java.util.function.Function;
+
```
nahsra, line 9: JavaDoc

nahsra (approved):
thig

PR #390, opened 2024-06-20
✨ New Sonar Remediation Codemod for Unsafe Reflection

gilday:
Remediates Sonar findings of type java:S2658.

Does not yet cover edge-cases, but I'd like to get some feedback on the happy path while I work on those edge cases.

/towards #work


framework/codemodder-testutils/src/main/java/io/codemodder/testutils/CodemodTestMixin.java
```diff
@@ -296,9 +298,7 @@ private List<Path> buildSonarJsonPaths(
    */
   default void verifyTransformedCode(final Path before, final Path expected, final Path after)
       throws IOException {
-    String expectedCode = Files.readString(expected);
-    String transformedJavaCode = Files.readString(after);
-    assertThat(transformedJavaCode, equalTo(expectedCode));
+    Assertions.assertThat(after).hasSameTextualContentAs(expected, StandardCharsets.UTF_8);
```
gilday, line 301: Getting much better diff in the test failure message with this.

core-codemods/src/main/java/io/codemodder/codemods/UnsafeReflectionRemediationCodemod.java
```diff
@@ -0,0 +1,114 @@
...
+import java.util.List;
+import javax.inject.Inject;
+
+/** Sonar remediation codemod for S2658: Classes should not be loaded dynamically. */
+@Codemod(
+    id = "sonar:java/unsafe-reflection-s2658",
+    reviewGuidance = ReviewGuidance.MERGE_WITHOUT_REVIEW,
+    importance = Importance.HIGH,
+    executionPriority = CodemodExecutionPriority.HIGH)
+public final class UnsafeReflectionRemediationCodemod
```
nahsra, line 25: Ditch `SonarPluginJavaParserChanger` in favor of the new `SonarRemediatingJavaParserChanger`. The new pattern makes it much easier to report CodeTF stuff that's important for where we're headed (aligning fixes to tool findings downstream.) See `SonarSQLInjectionCodemod for an example`. 

All we lose is a little boilerplate, but we get much better reporting outcomes.

framework/codemodder-testutils/src/main/java/io/codemodder/testutils/CodemodTestMixin.java
```diff
@@ -296,9 +298,7 @@ private List<Path> buildSonarJsonPaths(
    */
   default void verifyTransformedCode(final Path before, final Path expected, final Path after)
       throws IOException {
-    String expectedCode = Files.readString(expected);
-    String transformedJavaCode = Files.readString(after);
-    assertThat(transformedJavaCode, equalTo(expectedCode));
```
nahsra, line 301: Ironically that's why I was using hamcrest -- the assertj APIs I was using were not giving me nice IntelliJ "Click to see.." diffs. Does this fix that?

nahsra (changes requested):

framework/codemodder-testutils/src/main/java/io/codemodder/testutils/CodemodTestMixin.java
gilday: I'm not getting IntelliJ's "Click to see diff" in neither Hamcrest nor AssertJ. When you saw that in IntelliJ, were you using the IntelliJ JUnit runner (vs the "delegate to Gradle" runner)? fwiw I tried [fixing this feature for the Gradle runner in IntelliJ last year](https://github.com/JetBrains/intellij-community/commit/b9c612ac25618478c6d29813ef4c2446eb47f6c7), but I guess the fix was brittle, because it's no longer working for me.

Without that IntelliJ feature helping out, the Hamcrest errors are unintelligible. AssertJ has the diffing built-in, so I don't need to rely on the IntelliJ regex finding its match.

nahsra:
I'm picking up this PR.

nahsra (approved):

PR #394, opened 2024-06-21
Remove explicit setEntityExpansion calls

nahsra:
When fixing XXE, users may find it helpful to also remove explicit turning on off entity expansion.


sip49 (approved):
LGTM

PR #398, opened 2024-06-24
⬆️ Upgrade GPT 3.5 Turbo Model

gilday:
gpt-3.5-turbo-0613 has been removed, so replace it with new models.

* Updated failed authentication logging codemod to use new models and `SarifToLLMForMultiOutcomeCodemod`.
* Added a `Model` type to describe GPT models and colocates model-specific logic such as token counting.

/close #work


plugins/codemodder-plugin-llm/src/main/java/io/codemodder/plugins/llm/Model.java
gilday, line 1: Added this interface in addition to the `StandardModel` enum, so that downstream projects may add their own model if needed (vs being constrained to the enum definition in `codemodder-java`).

core-codemods/src/test/resources/logback-test.xml
gilday, line 1: This helped me understand why tests were failing.

nahsra (approved):
This should help  a lot!

gilday:
nvm, I'm still getting flakey tests. I need to keep working on this.

plugins/codemodder-plugin-llm/src/main/java/io/codemodder/plugins/llm/SarifToLLMForBinaryVerificationAndFixingCodemod.java
```diff
@@ -191,10 +194,11 @@ private <T> T getLLMResponse(
             .functions(functionExecutor.getFunctions())
             .functionCall(ChatCompletionRequestFunctionCall.of(function.getName()))
             .temperature(temperature)
```
nahsra, line 196: I know we're kind of throwing darts here -- I'd throw this change out if it's not obviously helping, as it could affect other codemods performing well already.

nahsra (changes requested):
Consider `top_p` change before merging.

core-codemods/src/test/resources/log-failed-login/vulnerable/SaltedHashLoginModule.java.after
gilday, line 1: I changed this from "safe" to "vulnerable". I don't see a good reason why we shouldn't add the log statement here. It's the only place that has enough context to include the username in the audit log.

core-codemods/src/test/resources/log-failed-login/vulnerable/MainFrame.java.after
```diff
@@ -154,7 +154,7 @@ ResultSet rs = stmt.executeQuery(sql);
                     this.dispose();
             }
             else{
-                    Logger.getLogger(MainFrame.class.getName()).log(Level.WARNING, "Failed login attempt for user: " + uname1);
+                    Logger.getLogger(MainFrame.class.getName()).log(Level.INFO, "Failed login attempt for user: " + uname1);
```
gilday, line 157: We instruct the model to use the INFO level, so I don't know why this test insisted on WARNING level.

core-codemods/src/test/java/io/codemodder/codemods/LogFailedLoginCodemodTest.java
```diff
@@ -3,14 +3,45 @@
...
+ *   <dt>unsafe/SaltedHashLoginModule
+ *   <dd>lacks a log statement before returning the authenticated decision. That is the correct
+ *       place to log, because it has the username in scope.
+ */
 @Metadata(
     codemodType = LogFailedLoginCodemod.class,
     testResourceDir = "log-failed-login",
     dependencies = {})
 @OpenAIIntegrationTest
-@Disabled("codemod is in disrepair")
```
gilday, line 13: no amount of temperature and top_p changes are good enough to make this test stop flaking. I'm going to keep it disabled.

nahsra (approved):

PR #402, opened 2024-06-25
✅ Add OpenAI Integration Tests to GitHub Actions Workflow

gilday:
* Includes a backstop to ensure that we don't forget to configure these tests in CI again.
* New JUnit extension `OpenAIIntegrationTest` makes the logic for conditionally running tests more cohesive. Companion Gradle plugin sets the required environment variable from a Gradle property, to allow for flexible configuration in developer environments. Updated README accordingly.
* Migrate LLM test utilities to use Gradle java test fixtures (by ways of a new convention plugin).
* Disable tests for codemods that have fallen into disrepair.

/close #work


nahsra (approved):

gradle/build-plugins/src/main/kotlin/io.codemodder.java-test-fixtures.gradle.kts
```diff
@@ -0,0 +1,3 @@
+plugins {
+    `java-test-fixtures`
```
ryandens, line 2: personally i would just inline this rather than have a convention plugin - if we need common conventions for all codemodder test fixtures we can introduce this later

ryandens (approved):

gradle/build-plugins/src/main/kotlin/io.codemodder.java-test-fixtures.gradle.kts
gilday: The convention plugin feels worthwhile for consistency to help developers approaching this project. If I were to inline this, I believe it would be the only example of a Gradle plugin that we reference directly from a build script vs via a convention plugin.

PR #403, opened 2024-06-26
🐛 Reimplement Sensitive Logging Codemod

nahsra:
This change rebuilds the codemod with gpt-4o with another strategy that merges with JavaParser. It consistently produces the results needed right now.

/towards ISS-1496


core-codemods/src/main/java/io/codemodder/codemods/SensitiveDataLoggingCodemod.java
```diff
@@ -1,53 +1,205 @@
...
+            .model("gpt-4o-2024-05-13")
+            //            .model("gpt-4-0613")
+            .n(1)
+            .messages(List.of(new ChatMessage(ChatMessageRole.USER.value(), prompt)))
+            .build();
+    ChatCompletionResult completion = service.createChatCompletion(request);
+    ChatCompletionChoice response = completion.getChoices().get(0);
+    String responseText = response.getMessage().getContent();
+    if (responseText.startsWith("```json") && responseText.endsWith("```")) {
+      responseText = responseText.substring(7, responseText.length() - 3);
```
gilday, line 117: For clarity

```suggestion
      responseText = responseText.substring("```json".length(), responseText.length() - "```".length());
```

core-codemods/src/main/java/io/codemodder/codemods/SensitiveDataLoggingCodemod.java
```diff
@@ -1,53 +1,205 @@
...
+    List<Result> results = sarif.getResultsByLocationPath(context.path());
+
+    String rawFile;
+    try {
+      rawFile = Files.readString(context.path());
+    } catch (IOException e) {
+      throw new UncheckedIOException("Couldn't read file", e);
+    }
+
+    List<String> rawLines = rawFile.lines().toList();
```
gilday, line 56: Avoid the intermediate string to save memory.

```suggestion
    final List<String> rawLines;
    try (final Stream<String> lines = Files.lines(context.path())) {
      rawLines = lines.toList();
    } catch (IOException e) {
      throw new UncheckedIOException("Couldn't read file", e);
    }
```

Also, consider combining the code that adds the line numbers with this code that streams the lines from disk, to avoid holding on to a copy of `rawLines` without line numbers that is never used.

core-codemods/src/main/java/io/codemodder/codemods/SensitiveDataLoggingCodemod.java
```diff
@@ -1,53 +1,205 @@
...
+        }
+      }
+    }
+    return CodemodFileScanningResult.from(changes, List.of());
   }
 
-  @Override
-  protected String getFixPrompt() {
-    return getClassResourceAsString(getClass(), "fix_prompt.txt");
+  private SensitivityAndFixAnalysis performSensitivityAnalysis(
```
gilday, line 87: I believe this will work with log statements that span multiple lines, but consider adding a test case for that, since this implementation is sensitive to line numbers.

core-codemods/src/main/java/io/codemodder/codemods/SensitiveDataLoggingCodemod.java
```diff
@@ -1,53 +1,205 @@
...
+      final List<String> lines, final Integer startLine) throws IOException {
+    String codeSnippet = numberedContextWithExtraLines(lines, startLine, 10, 10);
```
gilday, line 89: What are these constants and what are the "extra lines"?

core-codemods/src/main/java/io/codemodder/codemods/SensitiveDataLoggingCodemod.java
```diff
@@ -1,53 +1,205 @@
...
+      }
+    }
+    return CodemodFileScanningResult.from(changes, List.of());
   }
 
-  @Override
-  protected String getFixPrompt() {
-    return getClassResourceAsString(getClass(), "fix_prompt.txt");
+  private SensitivityAndFixAnalysis performSensitivityAnalysis(
+      final List<String> lines, final Integer startLine) throws IOException {
```
gilday, line 88: Can we use a primitive here and move the `null` check closer to where we read the potentially `null` startLine from SARIF?

core-codemods/src/main/java/io/codemodder/codemods/SensitiveDataLoggingCodemod.java
```diff
@@ -1,53 +1,205 @@
...
+          }
+        }
+      }
+    }
+    return CodemodFileScanningResult.from(changes, List.of());
   }
 
-  @Override
-  protected String getFixPrompt() {
-    return getClassResourceAsString(getClass(), "fix_prompt.txt");
```
gilday, line 39: Let's delete this unused resource.

core-codemods/src/main/java/io/codemodder/codemods/SensitiveDataLoggingCodemod.java
```diff
@@ -1,53 +1,205 @@
...
-    super(sarif, openAI);
+      final OpenAIService openAIService) {
+    this.sarif = sarif;
+    this.service = Objects.requireNonNull(openAIService);
   }
 
   @Override
-  protected String getThreatPrompt(
-      final CodemodInvocationContext context, final List<Result> results) {
-    return getClassResourceAsString(getClass(), "threat_prompt.txt");
```
gilday, line 34: Let's delete this unused resource.

core-codemods/src/main/java/io/codemodder/codemods/SensitiveDataLoggingCodemod.java
```diff
@@ -1,53 +1,205 @@
...
+    /** The new statement with which to replace the old. */
+    String newStatement();
+  }
+
+  private static class SensitivityAndFixAnalysisDTO implements SensitivityAndFixAnalysis {
+    @JsonProperty("sensitive_analysis_text")
+    private String sensitiveAnalysisText;
+
+    @JsonProperty("is_data_directly_logged")
+    private String isDataDirectlyLogged;
```
gilday, line 159: This part of the model's response is unused by the codemod logic. Regardless, does this exist solely to help the model "reason"?

core-codemods/src/main/java/io/codemodder/codemods/SensitiveDataLoggingCodemod.java
```diff
@@ -1,53 +1,205 @@
...
+            .filter(s -> s.getRange().get().begin.line == line)
+            .toList();
+    if (statements.size() == 1) {
+      return Optional.of(statements.get(0));
+    }
+    return Optional.empty();
+  }
+
+  /** The results of the sensitivity analysis and, optionally, the fix to apply. */
+  private interface SensitivityAndFixAnalysis {
```
gilday, line 139: Does this interface exist to help distinguish the parts of the response should be made available to the codemod logic vs the parts of the response that exist to help get better responses from the model? If so, can we add comments to this effect?

core-codemods/src/main/java/io/codemodder/codemods/SensitiveDataLoggingCodemod.java
```diff
@@ -1,53 +1,205 @@
...
+    String isSensitiveAnalysisText();
+
+    /** Whether the statement logs sensitive data. */
+    boolean isSensitiveAndDirectlyLogged();
+
+    /** The new statement with which to replace the old. */
+    String newStatement();
+  }
+
+  private static class SensitivityAndFixAnalysisDTO implements SensitivityAndFixAnalysis {
```
gilday, line 154: consider a `record` to reduce boilerplate and get `equals`, `hashcode`, and more importantly `toString()` implementations

core-codemods/src/main/java/io/codemodder/codemods/SensitiveDataLoggingCodemod.java
```diff
@@ -1,53 +1,205 @@
...
+            .n(1)
+            .messages(List.of(new ChatMessage(ChatMessageRole.USER.value(), prompt)))
+            .build();
+    ChatCompletionResult completion = service.createChatCompletion(request);
+    ChatCompletionChoice response = completion.getChoices().get(0);
+    String responseText = response.getMessage().getContent();
+    if (responseText.startsWith("```json") && responseText.endsWith("```")) {
+      responseText = responseText.substring(7, responseText.length() - 3);
+    }
+    return new ObjectMapper().readValue(responseText, SensitivityAndFixAnalysisDTO.class);
```
gilday, line 119: We can reuse this relatively expensive `ObjectMapper` by initializing it in the constructor. Furthermore, we can create an `ObjectReader` specifically for `SensitivityAndFixAnalysisDTO` to further optimize this.

core-codemods/src/main/java/io/codemodder/codemods/SensitiveDataLoggingCodemod.java
```diff
@@ -1,53 +1,205 @@
...
+              is_it_sensitive_and_directly_logged: a boolean dictating whether it is sensitive and definitely and directly logged
+              new_line_to_replace: if sensitive and directly logged, the statement on line %d that should replace it -- remember to correctly JSON escape this value
+              """
+            .formatted(startLine, codeSnippet, startLine);
+
+    ChatCompletionRequest request =
+        ChatCompletionRequest.builder()
+            .temperature(0D)
+            .model("gpt-4o-2024-05-13")
+            //            .model("gpt-4-0613")
```
gilday, line 109: Let's remove the commented out model.

core-codemods/src/main/java/io/codemodder/codemods/SensitiveDataLoggingCodemod.java
```diff
@@ -1,53 +1,205 @@
...
+
+              sensitive_analysis_text: a careful, thorough analysis of whether the data is sensitive (specifically a password, session ID, security token, SSN, etc -- not a username)
+              is_data_directly_logged: a careful, thorough analysis of whether the data is definitely and directly logged (e.g., not just passed to another method inside to the scope, unless that's a method that obviously returns the given string)
+              is_it_sensitive_and_directly_logged: a boolean dictating whether it is sensitive and definitely and directly logged
+              new_line_to_replace: if sensitive and directly logged, the statement on line %d that should replace it -- remember to correctly JSON escape this value
+              """
+            .formatted(startLine, codeSnippet, startLine);
+
+    ChatCompletionRequest request =
+        ChatCompletionRequest.builder()
```
gilday, line 106: This API call does not use the function calling parameters like others do. Why is that?

core-codemods/src/main/java/io/codemodder/codemods/SensitiveDataLoggingCodemod.java
```diff
@@ -1,53 +1,205 @@
...
-  @Override
-  protected boolean isPatchExpected(Patch<String> patch) {
-    // This codemod should only delete lines.
-    for (AbstractDelta<String> delta : patch.getDeltas()) {
-      if (!(delta instanceof DeleteDelta<String>)) {
-        return false;
-      }
+  /**
+   * We can fix if there's only one statement on the given line (meaning, it may span multiple
+   * lines, but only one statement is started on the lin.
```
gilday, line 124: Screenshot 2024-06-26 at 10 53 12 AM

gilday (changes requested):
I'm getting flaky tests when I remove the `@Disabled` annotation from the associated test class. Is it your intention to keep the test disabled?

core-codemods/src/main/java/io/codemodder/codemods/SensitiveDataLoggingCodemod.java
gilday: > Also, consider combining the code that adds the line numbers with this code that streams the lines from disk, to avoid holding on to a copy of rawLines without line numbers that is never used.

Disregard. I see now that the intention is to read the lines from disk once then take snippets from it.

core-codemods/src/main/java/io/codemodder/codemods/SensitiveDataLoggingCodemod.java
gilday: I understand now

core-codemods/src/main/java/io/codemodder/codemods/SensitiveDataLoggingCodemod.java
gilday: Added

gilday (approved):

PR #404, opened 2024-06-26
Added support for changing Statement object in assignments.

andrecsilva:
The SQL parameterizer codemod will now be able to apply the fix when the statement object originates from an assignment, e.g.:
```java
Statement stmt = null
try{
stmt = conn.createStatement();
stmt.executeQuery(...);
}
...
```
This unfortunately comes at the cost of breaking code whenever the assignment doesn't shadow the initiation. Since this particular case requires more sophisticated tools (mainly flow analysis), it is restricted for single assignments and the other cases stays as future work for now.

### DIFF
--- a/core-codemods/src/test/resources/sql-parameterizer/Test.java.after
+++ b/core-codemods/src/test/resources/sql-parameterizer/Test.java.after
@@ -81,4 +81,17 @@ public final class Test {
     var rs = stmt.execute();
     return rs;
   }
+
+  public ResultSet singeAssigned(String input) throws SQLException {
+    String sql = "SELECT * FROM USERS WHERE USER = ?";
+    Statement stmt = conn.createStatement();
+    try {
+      stmt = (Statement) conn.prepareStatement(sql);
+      ((PreparedStatement) stmt).setString(1, input);
+      ResultSet rs = ((PreparedStatement) stmt).execute();
+      return rs;
+    } catch (Exception e) {
+    }
+  }
+
 }
--- a/core-codemods/src/test/resources/sql-parameterizer/Test.java.after
+++ b/core-codemods/src/test/resources/sql-parameterizer/Test.java.after
@@ -82,13 +82,13 @@ public final class Test {
     return rs;
   }
 
-  public ResultSet singeAssigned(String input) throws SQLException {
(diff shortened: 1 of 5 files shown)


core-codemods/src/test/resources/sql-parameterizer/Test.java.before
```diff
@@ -66,4 +66,16 @@ public final class Test {
     var rs = stmt.executeQuery(sql);
     return rs;
   }
+
```
nahsra, line 69: Typo: `singeAssigned`

core-codemods/src/test/resources/sql-parameterizer/Test.java.after
```diff
@@ -81,4 +81,17 @@ public final class Test {
     var rs = stmt.execute();
     return rs;
   }
```
nahsra, line 83: Tagging @ryandens for a second opinion. I think I wouldn't want to receive this change. It's too confusing, it's probably causing compiler warnings, etc.

core-codemods/src/test/resources/sql-parameterizer/Test.java.before
```diff
@@ -66,4 +66,16 @@ public final class Test {
...
+  public ResultSet singeAssigned(String input) throws SQLException {
+    String sql = "SELECT * FROM USERS WHERE USER = '" + input + "'";
+    Statement stmt = conn.createStatement();
+    try {
+      stmt = conn.createStatement();
+      ResultSet rs = stmt.executeQuery(sql);
+      return rs;
+    } catch (Exception e) {
+    }
+  }
```
nahsra, line 79: Is the reason this code can't cleanly become a `PreparedStatement` because the history of `input` isn't known, and it might have SQL code in it?

nahsra (changes requested):

core-codemods/src/test/resources/sql-parameterizer/Test.java.before
andrecsilva: Not really, the reason is simply because the changes are local and less prone to break something. For this restricted case, I can probably change it without breaking it. Will do so in the next commit.

nahsra (approved):

PR #405, opened 2024-06-27
🚨 Fix Sonar Warnings About Boxed Primitives

gilday:
Sonar correctly identifies that we should prefer specialized function types that use primitive ints.

/towards #work

### DIFF
--- a/framework/codemodder-base/src/main/java/io/codemodder/remediation/reflectioninjection/ReflectionInjectionRemediator.java
+++ b/framework/codemodder-base/src/main/java/io/codemodder/remediation/reflectioninjection/ReflectionInjectionRemediator.java
@@ -4,7 +4,9 @@ import com.github.javaparser.ast.CompilationUnit;
 import io.codemodder.CodemodFileScanningResult;
 import io.codemodder.codetf.DetectorRule;
 import java.util.List;
+import java.util.OptionalInt;
 import java.util.function.Function;
+import java.util.function.ToIntFunction;
 
 /** Remediates reflection injection vulnerabilities. */
 public interface ReflectionInjectionRemediator {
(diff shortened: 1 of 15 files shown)


framework/codemodder-base/src/main/java/io/codemodder/remediation/reflectioninjection/ReflectionInjectionRemediator.java
```diff
@@ -16,8 +18,26 @@ <T> CodemodFileScanningResult remediateAll(
       DetectorRule detectorRule,
       List<T> issuesForFile,
       Function<T, String> getKey,
-      Function<T, Integer> getLine,
-      Function<T, Integer> getColumn);
+      ToIntFunction<T> getLine,
+      Function<T, OptionalInt> getColumn);
+
+  default <T> CodemodFileScanningResult remediateAll(
```
gilday, line 24: This method is necessary, because the tests do not provide column data.

On one hand, I believe this makes the tests inaccurate. For example, one of the "unfixable" examples is unfixable because there are multiple statements on the same line; however, given column information, I believe this would actually be fixable.

On the other hand, tools vary in their interpretation of the column, so maybe it's wrong to pick a single interpretation for the test?

nahsra:
I rejected this suggestion originally by Sonar because `column` is intentionally `null`. Binary scanners and IAST tools won't provide it.

framework/codemodder-base/src/main/java/io/codemodder/remediation/reflectioninjection/ReflectionInjectionRemediator.java
nahsra: > On the other hand, tools vary in their interpretation of the column, so maybe it's wrong to pick a single interpretation for the test?

Yup. Column could be `null` or +- between tools. That's why IMO we definitely need tests for the individual tools to confirm behavior.

nahsra (approved):

gilday:
I pushed a change that makes the column retrieval strategy all or nothing: the codemod for the tool x rule combination either always supplies column data or never supplies it. It doesn't leave room for a codemod that _sometimes_ provides column data.

I think this assumption is too strong. I'm thinking a more flexible design that allows for a codemod to sometimes provide the column data and sometimes not provide it may be better. However, I don't have an example of a codemod that needs to sometimes pass column information.

I don't want to belabor this too much, but I think I finally understand what's happening here, so I'm going to make that change and be done with this.

nahsra (approved):

nahsra:
Got out of date, and we recently switched remediator to another interface.

PR #406, opened 2024-06-27
✅ Add Tests For Reflection Injection Edge Case

gilday:
Add a tests for the edge case when code invokes the Class.forName(String) method using the fully qualified class name.

/close #work


framework/codemodder-base/src/test/java/io/codemodder/remediation/reflectioninjection/DefaultReflectionInjectionRemediatorTest.java
```diff
@@ -28,60 +30,71 @@ void setup() {
     this.rule = new DetectorRule("reflection-injection", "Reflection Injection", null);
   }
 
-  private static Stream<Arguments> fixableSamples() {
```
gilday, line 31: I really needed the developer experience of being able to run a single test case with a name, and the ability to add JavaDoc and annotations to an individual test case. That's why I changed these parameterized tests to be a set of related test cases.

nahsra (approved):

PR #407, opened 2024-06-27
👷 Add OpenAI Key as Task Input

gilday:


nahsra (approved):

gradle/build-plugins/src/main/kotlin/io.codemodder.openai.gradle.kts
```diff
@@ -11,9 +10,10 @@ testing {
             targets {
                 all {
                     testTask {
-                        val openAIKey = project.findProperty("codemodderOpenAIKey") as String?
-                        openAIKey?.let {
-                            environment("CODEMODDER_OPENAI_API_KEY", it)
+                        val openAIKey = providers.gradleProperty("codemodderOpenAIKey")
+                        inputs.property("codemodderOpenAIKey", openAIKey)
```
ryandens, line 14: i'm surprised/skeptical this is necessary - gradle considers environment variables as inputs for this task already. Did you experience this task not being up to date when this property changed?

gilday: I became aware of `providers.gradleProperty`, and it seemed like the configuration-cache friendly way to reference a Gradle property. I don't think it makes a difference today, because

1. As far as I know, we're not yet using the configuration cache.
2. A change to any Gradle property invalidates the configuration cache, but [more fine-grained cache invalidation is a planned feature](https://github.com/gradle/gradle/issues/20969).

I used this logic in `pro-java-codemods`, so I back-ported it here for consistency and to have this _should we do this_ discussion.

I was surprised to learn that there was no `environment` overload that accepted a `Provider` for the value. Because this didn't exist, I had to do the `isPresent` then `get()` strategy. I assumed that would be a problem for tracing the value passed to `environment` back to an individual Gradle property (which, again, is not yet valuable), so I added the property as an input.

gradle/build-plugins/src/main/kotlin/io.codemodder.openai.gradle.kts
```diff
@@ -11,9 +10,10 @@ testing {
...
+                        if (openAIKey.isPresent) {
+                            environment("CODEMODDER_OPENAI_API_KEY", openAIKey.get())
```
ryandens, line 16:
```suggestion
                        val openAIKey = providers.gradleProperty("codemodderOpenAIKey")
                        inputs.property("codemodderOpenAIKey", openAIKey)
                        if (openAIKey.isPresent) {
                            environment("CODEMODDER_OPENAI_API_KEY", openAIKey)
```

In my experience, Gradle often is able to track this kind of data flow, but interestingly environment variables for Test tasks are a special case where even if those values change they are not considered different for up to date checks. You can verify this on `main` today by running the test suite with a valid OpenAI key, changing the property to a known fake one, and then rerun the build and see that although the configuration cache will be updated to consider the changed property, the task itself won't be rerun (which seems like bad behavior for our use case). 


I tracked down the relevant documentation explaining why this is the case: https://docs.gradle.org/current/userguide/common_caching_problems.html#environment_variable_tracking - your solution here to specify an input property is the recommended one for ensuring the task will not be up to date if this environment variable changes.  



Note, although there is no `environment` overload that accepts a `Provider`, it does accept `Any`. This is a bit of a hangover from when Gradle was more commonly used with the Groovy DSL (I think) and type safety was more of a guideline than a rule. In general, when a Gradle API accepts `Any` where its implied that a task input/property/string value is supposed to be supplied, Gradle will use runtime logic to map it to the appropriate value at runtime

By avoiding a property.get() call, we avoid resolving the property until later (or potentially not at all if the task is not in the task graph).

ryandens (approved):
minor change only and some interesting TIL's about Gradle test tasks

gradle/build-plugins/src/main/kotlin/io.codemodder.openai.gradle.kts
ryandens: The downside to this apporach is that it would not be ideal if we were to introduce an HTTP build cache, as everyone would have a different cache key so the OpenAI tests would never be considered up to date.

gradle/build-plugins/src/main/kotlin/io.codemodder.openai.gradle.kts
ryandens: Forgot to mention this:
1. we are using the configuration cache in this project: https://github.com/pixee/codemodder-java/blob/main/gradle.properties#L3

PR #408, opened 2024-06-27
Create and use shared SQL injection reporting metadata

nahsra:


core-codemods/src/main/java/io/codemodder/codemods/SonarSQLInjectionCodemod.java
```diff
@@ -26,7 +27,7 @@ public final class SonarSQLInjectionCodemod extends SonarRemediatingJavaParserCh
   @Inject
   public SonarSQLInjectionCodemod(
       @ProvidedSonarScan(ruleId = "java:S2077") final RuleHotspot hotspots) {
-    super(CodemodReporterStrategy.fromClasspath(SQLParameterizerCodemod.class), hotspots);
```
gilday, line 29: Should we also update this codemod to use this generic strategy, or is the generic strategy only for tools?

framework/codemodder-base/src/test/java/io/codemodder/remediation/GenericVulnerabilityReporterStrategiesTest.java
```diff
@@ -17,7 +17,8 @@ private static Stream<Arguments> reporters() {
         Arguments.of(GenericVulnerabilityReporterStrategies.xxeReporterStrategy),
         Arguments.of(GenericVulnerabilityReporterStrategies.jndiReporterStrategy),
         Arguments.of(GenericVulnerabilityReporterStrategies.headerInjectionReporterStrategy),
-        Arguments.of(GenericVulnerabilityReporterStrategies.reflectionInjectionStrategy));
+        Arguments.of(GenericVulnerabilityReporterStrategies.reflectionInjectionStrategy),
+        Arguments.of(GenericVulnerabilityReporterStrategies.sqlInjectionStrategy));
```
gilday, line 21: If you instead make `GenericVulnerabilityReporterStrategies` an enum, then you can use `EnumSource` with this parameterized test, so that it's not possible to forget one of the enum values in the parameterized test arguments.

Also, `Arguments` is not needed here.

gilday (approved):

core-codemods/src/main/java/io/codemodder/codemods/SonarSQLInjectionCodemod.java
nahsra: Yeah, this is only for tool remediation.

PR #418, opened 2024-07-11
Skip LLM codemods when no service is available

drdavella:


plugins/codemodder-plugin-llm/src/main/java/io/codemodder/plugins/llm/SarifPluginLLMCodemod.java
```diff
@@ -0,0 +1,19 @@
+package io.codemodder.plugins.llm;
+
+import io.codemodder.RuleSarif;
+import io.codemodder.SarifPluginRawFileChanger;
+import java.util.Objects;
+
```
nahsra, line 6: Would add a comment here when to use this type since it's public.

core-codemods/src/main/java/io/codemodder/codemods/DefaultCodemods.java
```diff
@@ -35,6 +35,7 @@ public static List<Class<? extends CodeChanger>> asList() {
         JDBCResourceLeakCodemod.class,
         JEXLInjectionCodemod.class,
         JSPScriptletXSSCodemod.class,
```
nahsra, line 37: Should this be commented out?

plugins/codemodder-plugin-llm/src/main/java/io/codemodder/plugins/llm/SarifPluginLLMCodemod.java
```diff
@@ -0,0 +1,19 @@
...
+public abstract class SarifPluginLLMCodemod extends SarifPluginRawFileChanger {
+  protected final OpenAIService openAI;
+
+  public SarifPluginLLMCodemod(RuleSarif sarif, final OpenAIService openAI) {
+    super(sarif);
+    this.openAI = Objects.requireNonNull(openAI);
+  }
+
+  @Override
```
nahsra, line 15: Would add a comment here that I think subclassses shouldn't override this without also considering the same logic?

core-codemods/src/main/java/io/codemodder/codemods/DefaultCodemods.java
drdavella: It's in the catalog downstream so I can delete it here. I needed it for testing and just left it in-place.

nahsra (approved):

PR #422, opened 2024-07-18
✅ Added Tests for EnvironmentBasedModelMapper

gilday:
While working to confirm that the logic in codemodder-java and the platform is in sync, I found opportunities to add missing tests and fix some pitfalls.

/towards #work

### DIFF
--- a/plugins/codemodder-plugin-llm/src/main/java/io/codemodder/plugins/llm/EnvironmentBasedModelMapper.java
+++ b/plugins/codemodder-plugin-llm/src/main/java/io/codemodder/plugins/llm/EnvironmentBasedModelMapper.java
@@ -1,22 +1,34 @@
 package io.codemodder.plugins.llm;
 
 import java.util.HashMap;
+import java.util.Map;
 
 /** Mapper that maps models to their deployment names based on environment variables. */
 final class EnvironmentBasedModelMapper implements ModelMapper {
-  private static final String DEPLOYMENT_TEMPLATE = "CODEMODDER_AZURE_OPENAI_%s_DEPLOYMENT";
 
   private final HashMap<Model, String> map = new HashMap<>();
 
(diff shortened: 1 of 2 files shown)


plugins/codemodder-plugin-llm/src/main/java/io/codemodder/plugins/llm/EnvironmentBasedModelMapper.java
```diff
@@ -1,22 +1,34 @@
...
   @Override
   public String getModelName(Model model) {
-    return map.get(model);
+    return map.getOrDefault(model, model.id());
   }
+
+  /** Converts a model ID to environment variable casing. */
+  private static String toEnvironmentVariableCase(String input) {
+    return input.toUpperCase().replace('-', '_').replace('.', '_');
+  }
```
gilday, line 31: We should prefer explicitly changing the model ID into the expected casing, compared to relying on the implementation of the enum type's `toString()` method.

plugins/codemodder-plugin-llm/src/main/java/io/codemodder/plugins/llm/EnvironmentBasedModelMapper.java
```diff
@@ -1,22 +1,34 @@
...
+      final var name = String.format(DEPLOYMENT_TEMPLATE, toEnvironmentVariableCase(model.id()));
+      final var deployment = environment.getOrDefault(name, model.id());
+      map.put(model, deployment);
     }
   }
 
   @Override
   public String getModelName(Model model) {
-    return map.get(model);
+    return map.getOrDefault(model, model.id());
```
gilday, line 25: The previous logic did not account for non-standard models.

nahsra (approved):

PR #428, opened 2024-07-23
:sparkles: Support multiple rule names for AppScan

ryandens:
- **:sparkles: support muitiple rule names in AppScan**
- **:bulb: Improve docs for AppScan getRule accessor**

### DIFF
--- a/plugins/codemodder-plugin-appscan/src/main/java/io/codemodder/providers/sarif/appscan/AppScanModule.java
+++ b/plugins/codemodder-plugin-appscan/src/main/java/io/codemodder/providers/sarif/appscan/AppScanModule.java
@@ -41,10 +41,27 @@ public final class AppScanModule extends AbstractModule {
               .findFirst();
 
       annotation.ifPresent(
-          providedAppScanScan ->
+          providedAppScanScan -> {
+            if (!providedAppScanScan.ruleName().isEmpty()) {
               bind(RuleSarif.class)
                   .annotatedWith(providedAppScanScan)
-                  .toInstance(map.getOrDefault(providedAppScanScan.ruleName(), RuleSarif.EMPTY)));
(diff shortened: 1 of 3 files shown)


plugins/codemodder-plugin-appscan/src/main/java/io/codemodder/providers/sarif/appscan/AppScanModule.java
```diff
@@ -41,10 +41,27 @@ protected void configure() {
...
-          providedAppScanScan ->
+          providedAppScanScan -> {
+            if (!providedAppScanScan.ruleName().isEmpty()) {
               bind(RuleSarif.class)
                   .annotatedWith(providedAppScanScan)
-                  .toInstance(map.getOrDefault(providedAppScanScan.ruleName(), RuleSarif.EMPTY)));
+                  .toInstance(map.getOrDefault(providedAppScanScan.ruleName(), RuleSarif.EMPTY));
+            } else if (providedAppScanScan.ruleNames().length > 0) {
+
+              RuleSarif ruleSarif = RuleSarif.EMPTY;
```
nahsra, line 51: Not sure I understand this logic. We loop through the names and match/break on the first one?

I was expecting this would look much more like "merge N results together" kind of change.

ryandens: Typically, we see the different names as a result of different kinds of scans (e.g. from source or from a binary), so I wouldn't expect to see a SARIF with results from multiple of the names specified on the annotation in the same SARIF file. 

In general, I thought we were trying to avoid merging SARIF at this stage?

nahsra: Ah, true. It will only be one or the other, you're right.

nahsra (approved):

PR #434, opened 2024-07-31
Fixes issue where multiple mixed type injections were not fixed

andrecsilva:


framework/codemodder-base/src/main/java/io/codemodder/remediation/sqlinjection/SQLInjectionFixComposer.java
```diff
@@ -0,0 +1,26 @@
+package io.codemodder.remediation.sqlinjection;
+
+import com.github.javaparser.ast.expr.MethodCallExpr;
+
```
nahsra, line 4: JavaDoc

framework/codemodder-base/src/main/java/io/codemodder/remediation/sqlinjection/SQLInjectionFixComposer.java
```diff
@@ -0,0 +1,26 @@
...
+public final class SQLInjectionFixComposer {
+
+  private SQLInjectionFixComposer() {}
+
```
nahsra, line 8: JavaDoc

framework/codemodder-base/src/main/java/io/codemodder/remediation/sqlinjection/SQLInjectionFixComposer.java
```diff
@@ -0,0 +1,26 @@
...
+    if (maybeFixed.isPresent()) {
+      SQLParameterizerWithCleanup.cleanup(maybeFixed.get());
+      SQLTableInjectionFilterTransform.findAndFix(maybeFixed.get());
+      return true;
+      // If not, try the table injection only
+    } else {
+      return SQLTableInjectionFilterTransform.findAndFix(methodCallExpr);
+    }
+  }
+
```
nahsra, line 21: JavaDoc

Should this type be an `interface` and the others be implementations?

nahsra (changes requested):

framework/codemodder-base/src/main/java/io/codemodder/remediation/sqlinjection/SQLInjectionFixComposer.java
andrecsilva: A bit hesitant on that since these keep changing.  Notice how the `checkAndFix` for `SQLParameterization` is slightly different (returns an `Optional`), but this difference is necessary for the compostion.

I guess we could have a generic find-and-fix-transform interface if you feel the remediators need a stronger contract.

nahsra (approved):

PR #438, opened 2024-08-16
Adds per codemod Includes/Excludes 

andrecsilva:
Codemods now have their own includes/excludes rules and a `supports` method.

Includes/Excludes rules are specified by `PathMatcher` patterns that are relative to the repository root.

Before running on a given file, codemods will now perform two checks: includes/excludes and `supports`. 

Codemod specific Includes/excludes check can be overridden by global configuration from the command line (i.e. with `--path-include` flag). 

`supports` checks the bare-minimum conditions that must be satisfied before the codemod attempts to run. For example, 
javaparser codemods will check if the file is a java file and sarif based codemods will check if there is any results for the file. 

Refactored some scattered file checks in codemods into the `supports` method. For example, the `VerbTamperingCodemod` will now check if the file is a `web.xml` file with the `supports` method instead of the `visit`. This has the benefit of avoiding the generation of empty codetf reports.


core-codemods/src/intTest/java/io/codemodder/integration/WebGoat822Test.java
```diff
@@ -26,6 +26,29 @@ final class WebGoat822Test extends GitRepositoryTest {
     super("https://github.com/WebGoat/WebGoat", "main", "e75cfbeb110e3d3a2ca3c8fee2754992d89c419d");
   }
 
+  private static final String testPathIncludes =
```
nahsra, line 29: Can we expose the actual values so both the tests and the CLI use the same values?

core-codemods/src/main/java/io/codemodder/codemods/DefaultCodemods.java
```diff
@@ -57,6 +57,7 @@ public static List<Class<? extends CodeChanger>> asList() {
         SimplifyRestControllerAnnotationsCodemod.class,
         SubstituteReplaceAllCodemod.class,
         SonarXXECodemod.class,
```
nahsra, line 59: Nice catch.

core-codemods/src/main/java/io/codemodder/codemods/VerbTamperingCodemod.java
```diff
@@ -107,4 +104,14 @@ public String getIndividualChangeDescription(final Path filePath, final CodemodC
...
     return reporter.getReferences().stream().map(u -> new CodeTFReference(u, u)).toList();
   }
+
+  @Override
+  public IncludesExcludesPattern getIncludesExcludesPattern() {
+    return new IncludesExcludesPattern.Default(Set.of("{,**/}[wW][eE][bB].[xX][mM][lL]"), Set.of());
+  }
+
+  @Override
+  public boolean supports(Path file) {
```
nahsra, line 114: Make all these new args `final`

framework/codemodder-base/src/main/java/io/codemodder/CompositeJavaParserChanger.java
```diff
@@ -29,6 +30,17 @@ protected CompositeJavaParserChanger(final JavaParserChanger... changers) {
     this.changers = Arrays.asList(Objects.requireNonNull(changers));
   }
 
+  @Override
+  public IncludesExcludesPattern getIncludesExcludesPattern() {
```
nahsra, line 34: This doesn't seem right, but not sure I have another proposal right now.

framework/codemodder-base/src/main/java/io/codemodder/IncludesExcludesPattern.java
```diff
@@ -0,0 +1,77 @@
...
+   * and are treated as relative paths with regard to the repository root.
+   */
+  IncludesExcludes getRootedMatcher(final Path root);
+
+  class Default implements IncludesExcludesPattern {
+
+    private final Set<String> pathIncludes;
+    private final Set<String> pathExcludes;
+
+    public Default(final Set<String> pathIncludes, final Set<String> pathExcludes) {
```
nahsra, line 21: Can we make these null invariant?

E.g., `this.pathIncludes = Objects.requireNonNull(pathIncludes);`

framework/codemodder-base/src/main/java/io/codemodder/IncludesExcludesPattern.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+  }
+
+  /** An includes/excludes pattern that matches java files. */
+  final class JavaMatcherSingleton {
+    private static IncludesExcludesPattern singleton;
+
+    private JavaMatcherSingleton() {}
+
+    public static IncludesExcludesPattern getInstance() {
```
nahsra, line 44: Should this be `public`?

framework/codemodder-base/src/test/java/io/codemodder/CodemodLoaderTest.java
```diff
@@ -153,6 +163,11 @@ public List<CodeTFReference> getReferences() {
     public String getIndividualChangeDescription(Path filePath, CodemodChange change) {
       return null;
     }
+
+    @Override
```
nahsra, line 167: In the most root interface (if you wanted), you could provide a `default` interface method that returns true.

nahsra: To prevent all these simple implementations.

nahsra (changes requested):

framework/codemodder-base/src/test/java/io/codemodder/CodemodLoaderTest.java
andrecsilva: This was deliberate. Sarif and JavaParser based changers have reasonable defaults (sarif ones will check if the file is in the results and javaparser ones will check if it's a java file). 

Notice that all these are `RawFileChanger`s. Not having a default avoids creating codemods that inspect every file.

framework/codemodder-base/src/main/java/io/codemodder/CompositeJavaParserChanger.java
andrecsilva: Ideally this would be an intersection of all the patterns. But I'm not sure we can calculate this with just the patterns themselves.

core-codemods/src/intTest/java/io/codemodder/integration/WebGoat822Test.java
andrecsilva: CLI doesn't use these default includes anymore.

Do you still want these to be the default values for find-and-fix codemods? Right now every javaparser based one will only have `**.java` as its pattern.

framework/codemodder-base/src/main/java/io/codemodder/RawFileCodemodRunner.java
```diff
@@ -12,21 +12,27 @@
...
 
   private final RawFileChanger changer;
+  private final IncludesExcludes rootedFileMatcher;
 
-  RawFileCodemodRunner(final RawFileChanger changer) {
+  RawFileCodemodRunner(
+      final RawFileChanger changer,
+      final Path projectDir,
+      final IncludesExcludes globalIncludesExcludes) {
     this.changer = Objects.requireNonNull(changer);
```
nahsra, line 21: This is the only thing I don't like about this. This kind of subclass checking is a code smell of an architectural problem that will be hard to diagnose and fix later when it causes a problem.

nahsra (changes requested):

nahsra (approved):

PR #443, opened 2024-08-23
✨ Support CodeTF Findings Without IDs

gilday:
When a finding has no ID, we prefer to reflect this in CodeTF vs make up a new ID that is not significant to the tool that produced the finding.

/towards ISS-1837


gilday:
The issue identifier in my branch name is wrong, but the one in the PR description is correct. I didn't realize I had separate codemodder vs platform issues for this change.

nahsra (approved):

PR #449, opened 2024-09-24
New Sonar SSRF codemod

andrecsilva:
/close #work

Could not find any examples of sonar findings for SSRF vulnerabilities using `Url` or `HTTPUrlConnection`. This codemod only covers vulnerabilities found using the `RestTemplate` class from spring.


nahsra (changes requested):
We have a strategy for helping with this issue, even if it's not a complete resolution of all risk -- see  `SSRFCodemod`. This should use that same strategy. We may need to create an `SSRFRemediator`

core-codemods/src/main/java/io/codemodder/codemods/SonarSSRFCodemod.java
```diff
@@ -0,0 +1,58 @@
...
+import io.codemodder.remediation.GenericRemediationMetadata;
+import io.codemodder.sonar.model.Issue;
+import io.codemodder.sonar.model.SonarFinding;
+import java.util.List;
+import java.util.Objects;
+import javax.inject.Inject;
+
+/** Fixes SSRF issues found by sonsr rule javasecurity:S5144. */
+@Codemod(
+    id = "sonar:java/ssrf-s5144",
```
nahsra, line 19: Shouldn't this be more like cursory review? Isn't there zero effective risk of this change?

core-codemods/src/main/java/io/codemodder/codemods/remediators/ssrf/DefaultSSRFRemediator.java
```diff
@@ -45,8 +48,26 @@ public <T> CodemodFileScanningResult remediateAll(
...
-    FixCandidateSearchResults<T> results =
-        searcher.search(
+    // RestTemplate().exchange(url,...)
+    FixCandidateSearcher<T> rtSearcher =
+        new FixCandidateSearcher.Builder<T>()
+            // is method with name
+            .withMatcher(mce -> mce.isMethodCallWithName("exchange"))
+            // has RestTemplate as scope
+            .withMatcher(MethodOrConstructor::isMethodCallWithScope)
+            // The type check below doesn't work
```
nahsra, line 58: Let's pick one:

* fix the issue
* delete the comment and add a proper comment that this could be improved by further improving the type checking

We're not fans of commented code.

I must say I'm surprised the type checking doesn't resolve the type to either `RestTemplate` or the FQCN.

core-codemods/src/main/java/io/codemodder/codemods/remediators/ssrf/DefaultSSRFRemediator.java
```diff
@@ -45,8 +48,26 @@ public <T> CodemodFileScanningResult remediateAll(
             .withMatcher(mce -> !mce.getArguments().isEmpty())
             .build();
 
-    FixCandidateSearchResults<T> results =
-        searcher.search(
```
nahsra, line 49: I think it's not clear what this comment means, let's add a little more.

nahsra (changes requested):

core-codemods/src/main/java/io/codemodder/codemods/remediators/ssrf/DefaultSSRFRemediator.java
andrecsilva: It's an issue I've been observing with JavaParser's type resolver for quite some time. It works well with types from the Java API but struggles a bit with anything else.
I did a cursory investigation some time ago without much success. So, for now, I'll just comment it.

core-codemods/src/main/java/io/codemodder/codemods/SonarSSRFCodemod.java
andrecsilva: Changed it to match `SSRFCodemod`

nahsra (approved):

PR #455, opened 2024-10-11
Remediator API and adjacent classes refactoring

andrecsilva:
Adds a new API for the remediator/searcher pattern we use for codemods. The aim is to eliminate a lot of duplicate/similar-ish code around the remediators. Also converts the Sonar JNDI codemod to use the new API.

Classes from the previous API are still there, just renamed with the `Legacy` prefix.


nahsra:
I like the direction! No other comments. Can we use the strangler pattern to move remediators over, one at a time?

framework/codemodder-base/src/main/java/io/codemodder/remediation/SuccessOrReason.java
```diff
@@ -0,0 +1,54 @@
...
+import io.codemodder.Either;
+import java.util.List;
+import java.util.NoSuchElementException;
+import java.util.Objects;
+
+/**
+ * Represents the result of a strategy fix. Either a list of dependencies to be added, or a reason
+ * for failure.
+ */
+public class SuccessOrReason {
```
nahsra, line 13: Is this class meant to be extended? Make it `abstract`. If not, make it `final`. In truth, I would prefer an `interface` here. All `public` methods should also be JavaDoced.

framework/codemodder-base/src/main/java/io/codemodder/remediation/ModularRemediationStrategy.java
```diff
@@ -0,0 +1,20 @@
+package io.codemodder.remediation;
+
+import com.github.javaparser.ast.CompilationUnit;
+import com.github.javaparser.ast.Node;
+import java.util.function.BiFunction;
+
```
nahsra, line 6: `final` or `abstract`

nahsra: + JavaDoc

framework/codemodder-base/src/main/java/io/codemodder/remediation/DefaultLegacyFixCandidateSearcher.java
```diff
@@ -0,0 +1,123 @@
...
+import com.github.javaparser.ast.Node;
+import com.github.javaparser.ast.expr.MethodCallExpr;
+import com.github.javaparser.ast.expr.ObjectCreationExpr;
+import io.codemodder.codetf.DetectorRule;
+import io.codemodder.codetf.UnfixedFinding;
+import java.util.*;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import org.jetbrains.annotations.VisibleForTesting;
+
```
nahsra, line 14: Should we add a `@Deprecated` annotation here?

framework/codemodder-base/src/main/java/io/codemodder/remediation/jndiinjection/JNDIInjectionRemediator.java
```diff
@@ -1,31 +1,65 @@
...
 
-/**
- * Remediates JNDI injection vulnerabilities. It does this by weaving in a check to limit what JNDI
- * resources are available, and users can add more.
- *
- * <p>Inspiration for this came from logback:
- * https://github.com/qos-ch/logback/blob/979d76f3f2847f1c129bcc6295e69187d02e472c/logback-core/src/main/java/ch/qos/logback/core/util/JNDIUtil.java#L54
- */
-public interface JNDIInjectionRemediator {
+public class JNDIInjectionRemediator<T> implements Remediator<T> {
```
nahsra, line 12: These should still be interfaces with concrete implementations (e.g., `DefaultJNDIRemediator`) so they can be mocked for for testing. I know that we haven't invested much in testing the simple places in which they're used, but still...

framework/codemodder-base/src/main/java/io/codemodder/remediation/jndiinjection/JNDIFixContext.java
```diff
@@ -0,0 +1,86 @@
...
+    BlockStmt blockStmt,
+    Integer index) {
+
+  /**
+   * Test if a given node containing a JNDI vulnerability has all the relevant information for a
+   * fix.
+   *
+   * @param node
+   * @return Either the context with all the nodes for a fix, or a reason for failure.
+   */
```
nahsra, line 27: This method has around 10 exit locations, but is only tested indirectly as part of other components. Does it have enough coverage?

nahsra (changes requested):

framework/codemodder-base/src/main/java/io/codemodder/remediation/SuccessOrReason.java
andrecsilva: This class is mostly a wrapper for the `Either` value inside. The reason for this class is ease of use and expression. It should be pretty clear what it means instead of a generic `Either` type.

With that said, it is not meant be extended or behave in any other way other than the current implementation.

framework/codemodder-base/src/main/java/io/codemodder/remediation/jndiinjection/JNDIFixContext.java
andrecsilva: Some of them should be covered in the `DefaultJNDIRemediatorStrategyTest`, but I'm not sure every exit is covered.

framework/codemodder-base/src/main/java/io/codemodder/remediation/jndiinjection/JNDIInjectionRemediator.java
andrecsilva: Maybe I'm missing something, but I'm not sure I understand the point you're making entirely. 

The current class is the default implementation of the `Remediator` interface for JNDI. You could mock the `Remediator` if that's the problem.

I'm not sure creating a specific `JNDIInjectionRemediator` interface that is exactly, or largely, the same as the `Remediator` would accomplish.

Of course, I'm open to suggestions. Feel free to suggest a design that would accomplish what you mean.

nahsra (approved):

PR #462, opened 2024-10-24
Converted remaining remediators to new API and bugfixes

andrecsilva:


framework/codemodder-base/src/main/java/io/codemodder/remediation/FixCandidateSearcher.java
```diff
@@ -16,7 +16,7 @@ public interface FixCandidateSearcher<T> {
    * @param cu
    * @param path
    * @param rule
-   * @param issuesForFile
```
nahsra, line 19: It is a very cool idea to make this a mutable `List`. I wrestle with whether is a _good_ idea, because pure functions are easier to test and reason about. I am not inclined to stand in the way, just registering a concern.

framework/codemodder-base/src/main/java/io/codemodder/remediation/NodePositionMatcher.java
```diff
@@ -0,0 +1,50 @@
+package io.codemodder.remediation;
+
+import com.github.javaparser.ast.Node;
+
```
nahsra, line 4: JavaDoc

framework/codemodder-base/src/main/java/io/codemodder/remediation/WithoutScopePositionMatcher.java
```diff
@@ -0,0 +1,22 @@
...
+
+import com.github.javaparser.Position;
+import com.github.javaparser.Range;
+import com.github.javaparser.ast.Node;
+import com.github.javaparser.ast.nodeTypes.NodeWithTraversableScope;
+
+/** Removes the range of the node's scope before matching it against a position. */
+public final class WithoutScopePositionMatcher extends DefaultNodePositionMatcher {
+
+  @Override
```
nahsra, line 11: I think this type should be package-private, right? Does it need to be `public`? If it does, why is `getRange` not public?

framework/codemodder-base/src/main/java/io/codemodder/remediation/xxe/DocumentBuilderFactoryAndSAXParserAtCreationFixStrategy.java
```diff
@@ -0,0 +1,55 @@
...
+import com.github.javaparser.ast.Node;
+import com.github.javaparser.ast.body.VariableDeclarator;
+import com.github.javaparser.ast.expr.MethodCallExpr;
+import com.github.javaparser.ast.stmt.Statement;
+import io.codemodder.ast.ASTs;
+import io.codemodder.remediation.RemediationStrategy;
+import io.codemodder.remediation.SuccessOrReason;
+import java.util.List;
+import java.util.Optional;
+
```
nahsra, line 16: `final` + JavaDoc, and probably doesn't need to be `public`?

framework/codemodder-base/src/main/java/io/codemodder/remediation/xxe/SAXParserAtNewSPFixStrategy.java
```diff
@@ -0,0 +1,56 @@
...
+import com.github.javaparser.ast.Node;
+import com.github.javaparser.ast.expr.Expression;
+import com.github.javaparser.ast.expr.MethodCallExpr;
+import com.github.javaparser.ast.stmt.Statement;
+import io.codemodder.ast.ASTs;
+import io.codemodder.remediation.RemediationStrategy;
+import io.codemodder.remediation.SuccessOrReason;
+import java.util.List;
+import java.util.Optional;
+
```
nahsra, line 15: `final`

framework/codemodder-base/src/main/java/io/codemodder/remediation/xxe/XMLReaderAtParseFixStrategy.java
```diff
@@ -0,0 +1,71 @@
...
+        parser.asNameExpr(), parseStatement.get(), true);
+  }
+
+  /**
+   * Matches against XMLReader.parse calls
+   *
+   * @param node
+   * @return
+   */
+  public static boolean match(final Node node) {
```
nahsra, line 49: All of these sister types expose this static method -- this is a smell we may need a different architecture?

framework/codemodder-base/src/main/java/io/codemodder/remediation/xxe/XXEIntermediateXMLStreamReaderFixStrategy.java
```diff
@@ -0,0 +1,43 @@
...
+import com.github.javaparser.ast.CompilationUnit;
+import com.github.javaparser.ast.Node;
+import com.github.javaparser.ast.expr.Expression;
+import com.github.javaparser.ast.expr.MethodCallExpr;
+import com.github.javaparser.ast.expr.NameExpr;
+import com.github.javaparser.ast.stmt.Statement;
+import io.codemodder.remediation.RemediationStrategy;
+import io.codemodder.remediation.SuccessOrReason;
+import java.util.Optional;
+
```
nahsra, line 12: `final` and not `public` I don't think -- applies to all of these.

nahsra (changes requested):
Very cool! We're on the way. Before submitting a PR, make sure the easy stuff (`final`, `public`) etc. is right. 

I think there is a little more to the architecture to consider here, given the exposure of so many `public static` methods. Ideally we'd have no static methods that aren't generic utilities. Let's see if we can't iterate and make a solution that doesn't require them?

framework/codemodder-base/src/main/java/io/codemodder/remediation/DefaultNodePositionMatcher.java
```diff
@@ -0,0 +1,40 @@
+package io.codemodder.remediation;
+
+import com.github.javaparser.Position;
+import com.github.javaparser.Range;
+import com.github.javaparser.ast.Node;
+
```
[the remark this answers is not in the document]
andrecsilva: `WithoutScopePositionMatcher` extends it.

framework/codemodder-base/src/main/java/io/codemodder/remediation/WithoutScopePositionMatcher.java
andrecsilva: `XXERemediator` uses it as parameter. And thus, some codemods (Sonar based ones, mostly) make use of it. Also some tests.

`getRange` is there mostly to reuse code from `DefaultNodePositionMatcher` and shouldn't be accessed as it is an implementation detail. I've made it `protected`.

framework/codemodder-base/src/main/java/io/codemodder/remediation/FixCandidateSearcher.java
andrecsilva: This is admittedly, a bit of a hack since I've failed to accommodate for the problem multiple searchers reporting the same issues as unfixed before working on converting the XXE Remediators, where the problem became very apparent.

I don't think this is that much of a problem wince you shouldn't be using the `FixCandidateSearcher` outside of `SearcherStrategyRemediator` now. But I see your point.

I'll think about a solution for this.

framework/codemodder-base/src/main/java/io/codemodder/remediation/xxe/XMLReaderAtParseFixStrategy.java
andrecsilva: Each of those `match` methods should be their own `FixCandidateSearcher`, but it gets very verbose.

I'll think on an abstraction to make this easier.

nahsra (approved):

PR #468, opened 2024-11-17
Single codemod execution performance improvements

nahsra:
* Only load codemod resources if needed (a codemod for the provider is active)
* Cache AppScan location data


nahsra:
I closed and re-opened, after some investigation I was able to reduce the amount of changes shipped to get the intended effect.

drdavella (approved):

PR #480, opened 2024-12-03
Patched an issue with formatting empty elements

andrecsilva:
Fixed an issue where empty elements were not correctly being formatted.

### DIFF
--- a/plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/operator/FormatCommand.java
+++ b/plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/operator/FormatCommand.java
@@ -9,6 +9,7 @@ import java.io.StringWriter;
 import java.net.URISyntaxException;
 import java.nio.charset.Charset;
 import java.util.*;
+import java.util.function.UnaryOperator;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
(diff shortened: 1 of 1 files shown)


plugins/codemodder-plugin-maven/src/main/java/io/codemodder/plugins/maven/operator/FormatCommand.java
```diff
@@ -397,7 +416,7 @@ private StartElement getLastStartElement(List<XMLEvent> prevEvents) {
    * <p>this is important so we can mix and match offsets and apply formatting accordingly
    *
    * @param xmlDocumentString Rendered POM Document Contents (string-formatted)
-   * @return map of (index, matchData object) reverse ordered
+   * @return map of (index, matchData object) reve/rse ordered
```
nahsra, line 419: Typo

plugins/codemodder-plugin-maven/src/test/java/io/codemodder/plugins/maven/operator/POMOperatorTest.java
```diff
@@ -18,12 +18,15 @@
...
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.parallel.Execution;
+import org.junit.jupiter.api.parallel.ExecutionMode;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.xmlunit.diff.ComparisonType;
 import org.xmlunit.diff.Diff;
 import org.xmlunit.diff.Difference;
 
+@Execution(ExecutionMode.SAME_THREAD)
```
nahsra, line 29: This seems like a smell - do we need this and why?

nahsra (changes requested):

plugins/codemodder-plugin-maven/src/test/java/io/codemodder/plugins/maven/operator/POMOperatorTest.java
andrecsilva: No clue. Without this, the tests are flaky and interfere with each other. Sometimes it works, sometimes don't.

Seemed to work fine before adding the new test. But I did fix a bug with resource location when adding it.

nahsra: Yes please. I doubt the real use of this is multithreaded but I think at a minimum we need to understand this a little bit better and leave some comments or something.

andrecsilva: The issue was being caused because we were always using the same instance of `FormatCommand` from a static variable. I was able to reproduce it in the main branch, but only after dozens of runs.

nahsra (approved):

PR #484, opened 2024-12-10
New CodeQL codemod for mitigating Regex DoS 

andrecsilva:


nahsra (changes requested):
This is a complicated change that will be repeated many times in the codebase. Is there any chance we can get a function like this in the codebase, or add it to the java security toolkit?

If it’s something that must be delivered inline, we should have unit tests that test the code we insert to make sure it’s doing what we think it’s doing.

PR #489, opened 2025-01-09
Honor new combined Sonar flag

nahsra:
https://github.com/pixee/codemodder-specs/pull/43


drdavella (approved):
Noting that this is a breaking change since the old flags are no longer supported.

bdoyal (approved):

PR #494, opened 2025-02-07
Fixed error with changing execute statements in SQLi codemods

andrecsilva:


framework/codemodder-base/src/main/java/io/codemodder/remediation/sqlinjection/SQLParameterizer.java
```diff
@@ -725,7 +724,6 @@ private MethodCallExpr fixByHijackedStatement(
 
     // TODO will this work for every type of execute statement? or just executeQuery?
```
nahsra, line 725: Does this comment need to be adjusted?

nahsra (approved):

PR #496, opened 2025-02-19
Bump CodeTF bindings

nahsra:


drdavella (approved):

framework/codemodder-base/src/main/java/io/codemodder/DefaultCodemodExecutor.java
```diff
@@ -337,9 +337,17 @@ private FilesUpdateResult updateFiles(
 
     // create a changeset for this file change + its downstream dependency changes
     List<CodeTFChangesetEntry> changeset = new ArrayList<>();
+    Strategy codeChangeStrategy = codeTFAiMetadata != null ? Strategy.AI : Strategy.DETERMINISTIC;
```
drdavella, line 340: Does this affect every codemod? If so I can remove our logic for this in analysis service 🥳

nahsra: I can figure that out in a separate change if needed, but this should be broadly correct if AI metadata is being provided correctly.

